PR #1, opened 2019-09-10
[+] fix typo

122689305:
A typo in the path


dasmdasm:
That's not a type, that's how to refer to a build target in Bazel. See https://docs.bazel.build/versions/master/tutorial/cpp.html#build-the-project

PR #3, opened 2019-09-27
Enable syntax highlighting for code snippet.

johnjohndoe:


dasmdasm:
Good suggestion, thanks! We're still trying to get our tooling set up properly to accept pull requests, but we'll be happy to accept this once we have that working, hopefully in the next couple weeks.

Issue #24, opened 2020-07-16
[privacy-on-beam] Broken go.mod


bamnet:
Currently it's only possible to use the pbeam package via Bazel. The go.mod file is broken and generates errors when you `import github.com/google/differential-privacy/privacy-on-beam` into a codebase that uses native go modules.

```bash
$ go run main.go 
go: finding module for package github.com/google/differential-privacy/privacy-on-beam/pbeam
go: found github.com/google/differential-privacy/privacy-on-beam/pbeam in github.com/google/differential-privacy/privacy-on-beam v0.0.0-20200713153156-006ab1f8f903
go: github.com/google/differential-privacy/privacy-on-beam/pbeam: github.com/google/differential-privacy/privacy-on-beam@v0.0.0-20200713153156-006ab1f8f903: parsing go.mod: go.mod:8: require github.com/apache/beam: version "v2.22.0" invalid: should be v0 or v1, not v2
```

bamnet:
The fix here should be pretty straightforward - regenerating the go.mod by running the codelab (w/o bazel) should automatically update the syntax.  I can send a PR later today.

I _think_ this error is because `go.mod` is pulling in beam v2.22.0 without the +INCOMPATIBLE suffix.  As a result, Go is trying to resolve a [v2 style module](https://blog.golang.org/v2-go-modules), and beam is not using v2 style go modules.

bamnet:
In my tests, go can now correctly resolve the module and adds something like the following when import'ed.  Woohoo

```
	github.com/google/differential-privacy/go v0.0.0-20200723160438-d9f9215deec7 // indirect
	github.com/google/differential-privacy/privacy-on-beam v0.0.0-20200723160438-d9f9215deec7
```

PR #25, opened 2020-07-17
Update go.mod reference to beam.

bamnet:
This fixes the version reference to apache/beam which is not a "v2" style go package, closing #24.  It also apply standard Go formatting to go.mod.


miracvbasaran:
Thanks for the PR! :)

PR #31, opened 2020-07-28
Client-Specified Partitions for Privacy on Beam

r-jy:
Currently, the Privacy on Beam library uses thresholding to prevent distinguishing events from happening, which sometimes involves the loss of certain partitions. 

This PR adds functionality for clients to specify partitions (i.e., bins) to Privacy on Beam before running the pipeline, which would allow a client to keep all of their partitions. This means that the loss of some partitions through thresholding will be prevented if a client who uses the library specifies the partitions in a non data-dependent way.  Thus, in the common case where a client has partitions with a small number of users contributing to them, they can keep those partitions along with the data (thresholding, on the other hand, could result in the loss of this data). 

This PR also adds tests for those changes, including tests that check that the correct partitions are dropped and kept, and that dropped partitions are not contributed to (that dropped partitions do not subtract from the cross-partition contribution bound).


dibakch:
FYI: We've updated our default branch to the `main` branch.  Please update your PR.  Thanks! :)

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,213 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	switch vKind {
+	case reflect.Int64:
+		fn = prepareAddPartitionsInt64Fn
+	case reflect.Float64:
+		fn = prepareAddPartitionsFloat64Fn
+	default:
+		log.Exitf("pbeam.newPrepareAddPartitionsFn: vKind(%v) should be int64 or float64", vKind)
+	}
+
+	if err != nil {
```
johnSerrano, line 504: Can err ever not be nil here?

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,213 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	switch vKind {
+	case reflect.Int64:
+		fn = prepareAddPartitionsMeanInt64Fn
+	case reflect.Float64:
+		fn = prepareAddPartitionsMeanFloat64Fn
+	default:
+		log.Exitf("pbeam.newPrepareAddPartitionsFn: vKind(%v) should be int64 or float64", vKind)
+	}
+
+	if err != nil {
```
johnSerrano, line 523: Here as well - can err ever be set?

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,213 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	return partitionKey, []float64{}
+}
+
+
+type Partition struct {
+	PartitionKey string
+}
+
+// Function for sum and mean. Drop unspecified partitions.
+func correctPartitions(s beam.Scope, partitions []beam.PCollection, pcol PrivatePCollection, partitionT beam.EncodedType) beam.PCollection {
```
johnSerrano, line 551: can we rename this to "dropUnspecifiedPartitions" ?

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,213 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	if len(partitions) == 1 {
+		partitionsCol := partitions[0]
+		partitionsMap := beam.Combine(s, newKVPartitionsHashMapFn(partitionT), partitionsCol)
+		return beam.ParDo(s, prunePartitionsFn, pcol.col, beam.SideInput{Input: partitionsMap})
+	}
+	return pcol.col
+}
+
+// Function for count and distinct_id. Drop unspecified partitions.
+func correctCountPartitions(s beam.Scope, partitions []beam.PCollection, pcol PrivatePCollection, partitionT typex.FullType) beam.PCollection {
```
johnSerrano, line 561: Similarly here, can we call this something like DropUnspecifiedPartitionsForCount?

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,213 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	}
+	return b
+}
+
+func (fn *partitionsHashMapFn) ExtractOutput(m mapAccum) map[Partition]bool {
+	return m.PartitionMap
+}
+
+// Makes hashmaps for partitions stored in kv pairs.
+type KVpartitionsHashMapFn struct {
```
johnSerrano, line 615: Why do we need both of these hash map functions? Can't we enforce that our partition list is always the same format?

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,213 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	}
+	return b
+}
+
+func (fn *KVpartitionsHashMapFn) ExtractOutput(m mapAccum) map[Partition]bool {
+	return m.PartitionMap
+}
+
+
+// preparePruneFn takes a PCollection<K, V> as input, and returns a
```
johnSerrano, line 655: If this drops the partitions, why is "prepare" in the name?

privacy-on-beam/pbeam/count.go
```diff
@@ -68,37 +68,52 @@ type CountParams struct {
 // This aggregation is not hardened for such applications yet.
 //
 // Count transforms a PrivatePCollection<V> into a PCollection<V, int64>.
-func Count(s beam.Scope, pcol PrivatePCollection, params CountParams) beam.PCollection {
-	s = s.Scope("pbeam.Count")
+func Count(s beam.Scope, pcol PrivatePCollection, params CountParams, partitions ...beam.PCollection) beam.PCollection {
```
johnSerrano, line 71: Would be nice to update the function comment here to explain that you can only specify at most one partition

privacy-on-beam/pbeam/count.go
```diff
@@ -68,37 +68,52 @@ type CountParams struct {
...
+	if len(partitions) > 1 {
+		log.Exitf("Only one partition PCollection can be specified. %v were specified.", len(partitions))
+	}
+
 	// Obtain type information from the underlying PCollection<K,V>.
 	idT, partitionT := beam.ValidateKVType(pcol.col)
+	if len(partitions) == 1 {
```
johnSerrano, line 78: This seems to assume that all the specified partitions are the same type - can we be sure of this?

privacy-on-beam/pbeam/count.go
```diff
@@ -68,37 +68,52 @@ type CountParams struct {
...
 	epsilon, delta, err := spec.consumeBudget(params.Epsilon, params.Delta)
-	err = checkCountParams(params, epsilon, delta)
-	if err != nil {
-		log.Exit(err)
-	}
-
 	if err != nil {
 		log.Exitf("couldn't consume budget: %v", err)
 	}
+	err = checkCountParams(params, epsilon, delta)
```
johnSerrano, line 93: Why was this moved?

privacy-on-beam/pbeam/count.go
```diff
@@ -107,16 +122,39 @@ func Count(s beam.Scope, pcol PrivatePCollection, params CountParams) beam.PColl
...
 		countPairs,
 		beam.TypeDefinition{Var: beam.XType, T: partitionT.Type()})
+	return countWithCorrectPartitions(s, partitions, epsilon, delta, maxPartitionsContributed,
+		params, noiseKind, countsKV)
+
+}
+
+func countWithCorrectPartitions(s beam.Scope, partitions []beam.PCollection, epsilon, delta float64,
+	maxPartitionsContributed int64, params CountParams, noiseKind noise.Kind, countsKV beam.PCollection) beam.PCollection {
+	if len(partitions) == 0 { // No partitions specified.
```
johnSerrano, line 132: Is it possible to do this check in `count` and only call `countWithCorrectPartitions` if len > 0? Seems a little cleaner that way.

privacy-on-beam/pbeam/distinct_id.go
```diff
@@ -68,32 +72,39 @@ type DistinctPrivacyIDParams struct {
...
 	idT, partitionT := beam.ValidateKVType(pcol.col)
-
-	var noiseKind noise.Kind
-	if params.NoiseKind == nil {
-		noiseKind = noise.LaplaceNoise
-		log.Infof("No NoiseKind specified, using Laplace Noise by default.")
-	} else {
-		noiseKind = params.NoiseKind.toNoiseKind()
+	if len(partitions) == 1 {
+		if partitionT.Type() != partitions[0].Type().Type() {
```
johnSerrano, line 82: Same as above - can we assume all the elements of the partitions collection are the same type?

privacy-on-beam/pbeam/distinct_id.go
```diff
@@ -68,32 +72,39 @@ type DistinctPrivacyIDParams struct {
...
+
+	s = s.Scope("pbeam.DistinctPrivacyID")
+
 	// Get privacy parameters.
 	spec := pcol.privacySpec
 	epsilon, delta, err := spec.consumeBudget(params.Epsilon, params.Delta)
 	if err != nil {
 		log.Exitf("couldn't consume budget: %v", err)
 	}
-	err = checkDistinctPrivacyIDParams(params, noiseKind, epsilon, delta)
```
johnSerrano, line 89: Why was this deleted?

privacy-on-beam/pbeam/distinct_id.go
```diff
@@ -106,11 +117,28 @@ func DistinctPrivacyID(s beam.Scope, pcol PrivatePCollection, params DistinctPri
 	// done, remove the keys and count how many times each value appears.
 	values := beam.DropKey(s, decoded)
 	dummyCounts := beam.ParDo(s, addOneValueFn, values)
+
+	return distinctIdWithCorrectPartitions(s, partitions, epsilon, delta, maxPartitionsContributed, noiseKind, dummyCounts)
+}
+
+func distinctIdWithCorrectPartitions(s beam.Scope, partitions []beam.PCollection, epsilon, delta float64,
+	maxPartitionsContributed int64, noiseKind noise.Kind, dummyCounts beam.PCollection) beam.PCollection {
+	if len(partitions) == 0 {
```
johnSerrano, line 126: Similarly, it would be cleaner to do this before calling distinctIdWithCorrectPartitions

privacy-on-beam/pbeam/mean.go
```diff
@@ -89,8 +89,20 @@ type MeanParams struct {
...
-func MeanPerKey(s beam.Scope, pcol PrivatePCollection, params MeanParams) beam.PCollection {
+func MeanPerKey(s beam.Scope, pcol PrivatePCollection, params MeanParams, partitions ...beam.PCollection) beam.PCollection {
+	if len(partitions) > 1 {
+		log.Exitf("Only one partition PCollection can be specified. %v were specified.", len(partitions))
+	}
+
 	s = s.Scope("pbeam.MeanPerKey")
+
+	if len(partitions) == 1 {
+		if pcol.codec.KType.T != partitions[0].Type().Type() {
```
johnSerrano, line 100: Also here - can we assume the elements are all the same type?

privacy-on-beam/pbeam/mean.go
```diff
@@ -158,12 +174,34 @@ func MeanPerKey(s beam.Scope, pcol PrivatePCollection, params MeanParams) beam.P
 		partialPairs,
 		beam.TypeDefinition{Var: beam.XType, T: partitionT})
 
+	return meanWithCorrectPartitions(s, partitions, epsilon, delta, maxPartitionsContributed,
+		params, noiseKind, partialKV)
+}
+
+func meanWithCorrectPartitions(s beam.Scope, partitions []beam.PCollection, epsilon, delta float64,
+	maxPartitionsContributed int64, params MeanParams, noiseKind noise.Kind, partialKV beam.PCollection) beam.PCollection {
+	if len(partitions) == 0 { // no partitions specified
```
johnSerrano, line 183: Please move this out of this function if possible

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/google/differential-privacy/go/dpagg"
+	// "github.com/google/differential-privacy/go/dpagg"
```
johnSerrano, line 23: Delete?

privacy-on-beam/pbeam/sum.go
```diff
@@ -80,8 +80,19 @@ type SumParams struct {
...
 // input is an integer type or a float type.
-func SumPerKey(s beam.Scope, pcol PrivatePCollection, params SumParams) beam.PCollection {
+func SumPerKey(s beam.Scope, pcol PrivatePCollection, params SumParams, partitions ...beam.PCollection) beam.PCollection {
+	if len(partitions) > 1 {
+		log.Exitf("Only one partition PCollection can be specified. %v were specified.", len(partitions))
+	}
+
 	s = s.Scope("pbeam.SumPerKey")
+	if len(partitions) == 1 {
+		if pcol.codec.KType.T != partitions[0].Type().Type() {
```
johnSerrano, line 90: Also here, are types guaranteed to be consistent within a pcol?

privacy-on-beam/pbeam/sum.go
```diff
@@ -139,16 +156,39 @@ func SumPerKey(s beam.Scope, pcol PrivatePCollection, params SumParams) beam.PCo
...
 		partialSumPairs,
 		beam.TypeDefinition{Var: beam.XType, T: partitionT})
+	return sumWithCorrectPartitions(s, partitions, epsilon, delta, maxPartitionsContributed,
+		params, noiseKind, vKind, partialSumKV)
+
+}
+
+func sumWithCorrectPartitions(s beam.Scope, partitions []beam.PCollection, epsilon, delta float64,
+	maxPartitionsContributed int64, params SumParams, noiseKind noise.Kind, vKind reflect.Kind, partialSumKV beam.PCollection) beam.PCollection {
+	if len(partitions) == 0 { // No partitions specified.
```
johnSerrano, line 166: Please do this check outside of this function

privacy-on-beam/pbeam/aggregations.go
r-jy: I don't think so, because the PCollection data is formatted differently... @miracvbasaran ?

privacy-on-beam/pbeam/sum.go
r-jy: From what I know, pcol.codec.KType.T should be the type of all of the partitions in the original pcol. Are you suggesting that I check that the type of every specified partition is equal to pcol.codec.KType.T?

privacy-on-beam/pbeam/distinct_id.go
r-jy: It was shifted to line 97.

privacy-on-beam/pbeam/count.go
r-jy: Thought it would be more important to check that partitions are of the same type before checking the parameters of count.

r-jy: I think the previous file was wrong...

privacy-on-beam/pbeam/aggregations.go
r-jy: Just kidding, was able to just use one of them. Thanks John!

privacy-on-beam/pbeam/sum.go
r-jy: partitions[0] is the entire PCollection since partitions is input to the variadic function.

privacy-on-beam/pbeam/count.go
r-jy: I think the previous file was wrong...

privacy-on-beam/pbeam/mean.go
johnSerrano: ah, never mind, this isn't checking the first partition's type, but the pcollection's setting. I misread

privacy-on-beam/pbeam/sum.go
johnSerrano: ah makes sense. Thanks!

johnSerrano:
LGTM, thanks!

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,156 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+
+// Function for count and distinct_id. Drop unspecified partitions.
+func dropUnspecifiedPartitionsForCount(s beam.Scope, partitionsCol beam.PCollection, pcol PrivatePCollection, partitionFullType typex.FullType) beam.PCollection {
+	partitionEncodedType := beam.EncodedType{partitionFullType.Type()}
+	partitionMap := beam.Combine(s, newPartitionsHashMapFn(partitionEncodedType), partitionsCol)
+	return beam.ParDo(s, newPrunePartitionsFnForCount(partitionFullType), pcol.col, beam.SideInput{Input: partitionMap})
+}
+
+type mapAccum struct {
+	// Key is string representation of partition encoding.
```
miracvbasaran, line 550: nit: `Key is the string representation of the encoded partition key.`

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,156 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	// Value is true.
+	PartitionMap PMap
+}
+
+// Make hashmaps for partitions.
+type PartitionsHashMapFn struct {
```
miracvbasaran, line 556: nit: Maybe just "Map" instead of "HashMap" since that's the naming used in Go.

(Also in other places)

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,156 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+// Function for count and distinct_id. Drop unspecified partitions.
+func dropUnspecifiedPartitionsForCount(s beam.Scope, partitionsCol beam.PCollection, pcol PrivatePCollection, partitionFullType typex.FullType) beam.PCollection {
+	partitionEncodedType := beam.EncodedType{partitionFullType.Type()}
+	partitionMap := beam.Combine(s, newPartitionsHashMapFn(partitionEncodedType), partitionsCol)
+	return beam.ParDo(s, newPrunePartitionsFnForCount(partitionFullType), pcol.col, beam.SideInput{Input: partitionMap})
+}
+
+type mapAccum struct {
+	// Key is string representation of partition encoding.
+	// Value is true.
```
miracvbasaran, line 551: nit: "always true" or "always set to true"

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,156 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+}
+
+func (fn *PartitionsHashMapFn) ExtractOutput(m mapAccum) PMap {
+	return m.PartitionMap
+}
+
+// prunePartitionsFnForCount takes a PCollection<K, V> as input, and returns a
+// PCollection<K, V>, where unspecified partitions have been dropped.
+// Used for count and distinct_id.
+type prunePartitionsFnForCount struct {
```
miracvbasaran, line 599: The same naming suggestion here with `prunePartitionsVFn` and `prunePartitionsKVFn`

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,156 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+}
+
+func (fn *prunePartitionsFnForCount) Setup() {
+	fn.partitionEnc = beam.NewElementEncoder(fn.PartitionType.T)
+}
+
+func (fn *prunePartitionsFnForCount) ProcessElement(id beam.X, partitionKey beam.V, partitionsIter func(*PMap) bool, emit func(beam.X, beam.V)) {
+	var partitionBuf bytes.Buffer
+	if err := fn.partitionEnc.Encode(partitionKey, &partitionBuf); err != nil {
+		log.Exitf("pbeam.preparePruneFn.ProcessElement: couldn't encode partition %v: %v", partitionKey, err)
```
miracvbasaran, line 617: pbeam.prunePartitionsVFn

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +483,156 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	}
+	var partitionMap PMap
+	partitionsIter(&partitionMap)
```
miracvbasaran, line 620: Please check that partitionMap is not nil afterwards, also in the function below.

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -380,13 +391,35 @@ func (fn *boundedSumFloat64Fn) MergeAccumulators(a, b boundedSumAccumFloat64) bo
...
+	result := a.BS.Result()
+	if a.PartitionsSpecified || a.SP.Result() {
 		return &result
 	}
+
+	return nil
+}
+
+// Convert from *int64 to int64 or *float64 to float64
+func findCorrectToFn(kind reflect.Kind) interface{} {
```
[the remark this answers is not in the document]
r-jy: Should I rename the variable from kind to vKind?

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -380,13 +391,35 @@ func (fn *boundedSumFloat64Fn) MergeAccumulators(a, b boundedSumAccumFloat64) bo
...
+	switch kind {
+	case reflect.Int64:
+		return CorrectToInt64
+	case reflect.Float64:
+		return CorrectToFloat64
+	default:
+		log.Exitf("pbeam.findDropThresholdedPartitionsFn: kind(%v) should be int64 or float64", kind)
```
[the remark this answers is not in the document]
r-jy: Heh, thanks for catching this!

privacy-on-beam/pbeam/count_test.go
```diff
@@ -230,6 +264,38 @@ func TestCountCrossPartitionContributionBounding(t *testing.T) {
...
+	result := []testInt64Metric{
+		{0, 150},
+	}
+	p, s, col, want := ptest.CreateList2(pairs, result)
+	col = beam.ParDo(s, pairToKV, col)
+
+	partitions := []int{0, 1, 2}
+	partitionsCol := CreateList(s, partitions)
+
+	epsilon, delta, k, l1Sensitivity := 50.0, 0.01, 25.0, 3.0
```
miracvbasaran, line 283: Please add a comment about the parameters here.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -340,6 +475,40 @@ func TestSumPerKeyCrossPartitionContributionBoundingFloat(t *testing.T) {
...
+	}
+	result := []testFloat64Metric{
+		{0, 150.0},
+	}
+	p, s, col, want := ptest.CreateList2(triples, result)
+	col = beam.ParDo(s, extractIDFromTripleWithFloatValue, col)
+	partitions := []int{0, 1, 2}
+	partitionsCol := CreateList(s, partitions)
+
+	epsilon, delta, k, l1Sensitivity := 50.0, 0.01, 25.0, 3.0
```
miracvbasaran, line 494: Please add a comment about the parameters here.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -305,6 +408,38 @@ func TestSumPerKeyCrossPartitionContributionBoundingInt(t *testing.T) {
...
+	var triples []tripleWithIntValue
+	for i := 0; i < 10; i++ {
+		triples = append(triples, makeDummyTripleWithIntValue(50, i)...)
+	}
+	result := []testInt64Metric{
+		{0, 150},
+	}
+	p, s, col, want := ptest.CreateList2(triples, result)
+	col = beam.ParDo(s, extractIDFromTripleWithIntValue, col)
+	partitionsCol := CreateList(s, []int{0, 1, 2})
```
miracvbasaran, line 423: We are not actually testing cross-partition contribution bounding if we have a total of 3 kept partitions. We should keep more partitions and let Privacy on Beam drop the extra ones. So, let's say you keep `{0, 1, 2, 3, 4}`.

privacy-on-beam/pbeam/count_test.go
```diff
@@ -230,6 +264,38 @@ func TestCountCrossPartitionContributionBounding(t *testing.T) {
...
+	for i := 0; i < 10; i++ {
+		pairs = append(pairs, makePairsWithFixedV(50, i)...)
+	}
+	result := []testInt64Metric{
+		{0, 150},
+	}
+	p, s, col, want := ptest.CreateList2(pairs, result)
+	col = beam.ParDo(s, pairToKV, col)
+
+	partitions := []int{0, 1, 2}
```
miracvbasaran, line 280: We are not actually testing cross-partition contribution bounding if we have a total of 3 kept partitions. We should keep more partitions and let Privacy on Beam drop the extra ones. So, let's say you keep `{0, 1, 2, 3, 4}`.

privacy-on-beam/pbeam/distinct_id_test.go
```diff
@@ -275,6 +319,38 @@ func TestDistinctPrivacyIDCrossPartitionContributionBounding(t *testing.T) {
...
+	var pairs []pairII
+	for i := 0; i < 10; i++ {
+		pairs = append(pairs, makePairsWithFixedV(50, i)...)
+	}
+	result := []testInt64Metric{
+		{0, 150},
+	}
+	p, s, col, want := ptest.CreateList2(pairs, result)
+	col = beam.ParDo(s, pairToKV, col)
+	partitionsCol := CreateList(s, []int{0, 1, 2})
```
miracvbasaran, line 335: We are not actually testing cross-partition contribution bounding if we have a total of 3 kept partitions. We should keep more partitions and let Privacy on Beam drop the extra ones. So, let's say you keep `{0, 1, 2, 3, 4}`.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -340,6 +475,40 @@ func TestSumPerKeyCrossPartitionContributionBoundingFloat(t *testing.T) {
...
+	var triples []tripleWithFloatValue
+	for i := 0; i < 10; i++ {
+		triples = append(triples, makeDummyTripleWithFloatValue(50, i)...)
+	}
+	result := []testFloat64Metric{
+		{0, 150.0},
+	}
+	p, s, col, want := ptest.CreateList2(triples, result)
+	col = beam.ParDo(s, extractIDFromTripleWithFloatValue, col)
+	partitions := []int{0, 1, 2}
```
miracvbasaran, line 491: We are not actually testing cross-partition contribution bounding if we have a total of 3 kept partitions. We should keep more partitions and let Privacy on Beam drop the extra ones. So, let's say you keep `{0, 1, 2, 3, 4}`.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -305,6 +408,38 @@ func TestSumPerKeyCrossPartitionContributionBoundingInt(t *testing.T) {
...
+	}
+	p, s, col, want := ptest.CreateList2(triples, result)
+	col = beam.ParDo(s, extractIDFromTripleWithIntValue, col)
+	partitionsCol := CreateList(s, []int{0, 1, 2})
+
+	epsilon, delta, k, l1Sensitivity := 50.0, 0.01, 25.0, 3.0
+	pcol := MakePrivate(s, col, NewPrivacySpec(epsilon, delta))
+	pcol = ParDo(s, tripleWithIntValueToKV, pcol)
+	got := SumPerKey(s, pcol, SumParams{MaxPartitionsContributed: 3, MinValue: 0, MaxValue: 1, NoiseKind: LaplaceNoise{}, specifiedPartitions: partitionsCol})
+	// With a max contribution of 3, all of the data going to the three partitions
```
miracvbasaran, line 429: nit: "three partitions" instead of "the three partitions"

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -340,6 +475,40 @@ func TestSumPerKeyCrossPartitionContributionBoundingFloat(t *testing.T) {
...
+	p, s, col, want := ptest.CreateList2(triples, result)
+	col = beam.ParDo(s, extractIDFromTripleWithFloatValue, col)
+	partitions := []int{0, 1, 2}
+	partitionsCol := CreateList(s, partitions)
+
+	epsilon, delta, k, l1Sensitivity := 50.0, 0.01, 25.0, 3.0
+	pcol := MakePrivate(s, col, NewPrivacySpec(epsilon, delta))
+	pcol = ParDo(s, tripleWithFloatValueToKV, pcol)
+	got := SumPerKey(s, pcol, SumParams{MaxPartitionsContributed: 3, MinValue: 0.0, MaxValue: 1.0, NoiseKind: LaplaceNoise{}, specifiedPartitions: partitionsCol})
+	// With a max contribution of 3, all of the data for the three partitions should be kept.
```
miracvbasaran, line 498: nit: "three partitions" instead of "the three partitions"

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -154,6 +230,33 @@ func TestSumPerKeyNegativeBoundsFloat(t *testing.T) {
 	}
 }
 
+// Checks that SumPerKey with partitions works correctly for negative bounds and negative values with float values.
+func TestSumPerKeyWithPartitionsNegativeBoundsFloat(t *testing.T) {
```
miracvbasaran, line 234: Please add a test for integers as well.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -607,6 +776,35 @@ func TestSumPerKeyReturnsNonNegativeFloat64(t *testing.T) {
...
+	col = beam.ParDo(s, extractIDFromTripleWithFloatValue, col)
+
+	var partitions []int
+	for p := 0; p < 200; p++ {
+		partitions = append(partitions, p)
+	}
+	partitionsCol := CreateList(s, partitions)
+
+	// Using a low epsilon, a high delta, and a high maxValue here to add a
+	// lot of noise while keeping partitions.
```
miracvbasaran, line 795: nit: Remove "while keeping partitions." because we are not thresholding here.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -628,6 +826,34 @@ func TestSumPerKeyReturnsNonNegativeInt64(t *testing.T) {
...
+	col = beam.ParDo(s, extractIDFromTripleWithIntValue, col)
+
+	var partitions []int
+	for p := 0; p < 200; p++ {
+		partitions = append(partitions, p)
+	}
+	partitionsCol := CreateList(s, partitions)
+
+	// Using a low epsilon, a high delta, and a high maxValue here to add a
+	// lot of noise while keeping partitions.
```
miracvbasaran, line 845: nit: Remove "while keeping partitions." because we are not thresholding here.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -718,6 +718,47 @@ func TestMeanPerKeyReturnsNonNegativeFloat64(t *testing.T) {
...
+
+	var partitions []int
+	for p := 0; p < 200; p++ {
+		partitions = append(partitions, p)
+	}
+	partitionsCol := CreateList(s, partitions)
+
+	// Using a low ε, a high δ, and a high maxValue here to add a
+	// lot of noise while keeping partitions.
+	// ε=0.001. δ=0.999 and l0Sensitivity=1 gives a threshold of =2.
```
miracvbasaran, line 738: Remove comments about keeping partitions and threshold please.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -718,6 +718,47 @@ func TestMeanPerKeyReturnsNonNegativeFloat64(t *testing.T) {
...
+	maxContributionsPerPartition := int64(1)
+	epsilon := 0.001
+	delta := 0.999
+	lower := 0.0
+	upper := 1e8
+
+	// ε is split by 2 for noise and for partition selection, so we use 2*ε to get a Laplace noise with ε.
```
miracvbasaran, line 745: This comment made me realize a HUGE thing we forgot to address: We shouldn' spend budget on thresholding when we specify partitions.

Please add logic for that for all aggregations.

And please fix the comments and parameters for all tests with specified partitions.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -785,3 +826,128 @@ func TestFindConvertToFloat64Fn(t *testing.T) {
...
+
+	maxContributionsPerPartition := int64(1)
+	maxPartitionsContributed := int64(1)
+	epsilon := 60.0
+	delta := 0.01
+	lower := 0.0
+	upper := 150.0
+
+	// Specify partition 0, so id = 0 is guaranteed to contribute to partition 0.
+	partitionsCol := CreateList(s, []int{0})
```
miracvbasaran, line 862: Again, we don't test cross-partition contribution bounding if the number of specified partitions is equal to MaxPartitionsContributed. Please fix. (Also the comments)

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -785,3 +826,128 @@ func TestFindConvertToFloat64Fn(t *testing.T) {
...
+	if err := approxEqualsKVFloat64(s, got, want, tolerance1+tolerance2); err != nil {
+		t.Fatalf("TestMeanPerKeyWithPartitionsPerPartitionContributionBounding: %v", err)
+	}
+	if err := ptest.Run(p); err != nil {
+		t.Errorf("TestMeanPerKeyWithPartitionsPerPartitionContributionBounding: MeanPerKey(%v) = %v, want %v, error %v", col, got, want, err)
+	}
+}
+
+// Checks that MeanPerKey with specified partitions
+// returns a correct answer for int input values.
```
miracvbasaran, line 899: nit: COuld you format this so that the first line is "full" before moving to the second line?

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -785,3 +826,128 @@ func TestFindConvertToFloat64Fn(t *testing.T) {
...
+// They should be correctly converted to float64 and then correct result
+// with float statistic should be computed.
+func TestMeanPerKeyWithPartitionsNoNoiseIntValues(t *testing.T) {
```
miracvbasaran, line 902: Could you also add a test for Float values?

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -785,3 +826,128 @@ func TestFindConvertToFloat64Fn(t *testing.T) {
...
+	triples := concatenateTriplesWithIntValue(
+		makeTripleWithIntValue(7, 0, 2),
+		makeTripleWithIntValueStartingFromKey(7, 100, 1, 1),
+		makeTripleWithIntValueStartingFromKey(107, 150, 1, 2))
+
+	exactCount := 250.0
+	exactMean := (100.0 + 2.0*150.0) / exactCount
+
+	// ε=50, δ=10⁻²⁰⁰ and l0Sensitivity=1 gives a threshold of =11.
```
miracvbasaran, line 911: Please remove stuff about the threshold.

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +481,176 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+
+// dropUnspecifiedPartitionsKVFn drops partitions not specified in partitionsCol from pcol. It can be used for aggregations on <K,V> pairs, e.g. sum and mean.
+func dropUnspecifiedPartitionsKVFn(s beam.Scope, partitionsCol beam.PCollection, pcol PrivatePCollection, partitionEncodedType beam.EncodedType) beam.PCollection {
+	partitionMap := beam.Combine(s, newPartitionsMapFn(partitionEncodedType), partitionsCol)
+	return beam.ParDo(s, prunePartitionsKVFn, pcol.col, beam.SideInput{Input: partitionMap})
+
+}
+
+// dropUnspecifiedPartitionsVFn drops partitions not specified in partitionsCol from pcol. It can be used for aggregations on V values, e.g. count and distinctid.
+func dropUnspecifiedPartitionsVFn(s beam.Scope, partitionsCol beam.PCollection, pcol PrivatePCollection, partitionFullType typex.FullType) beam.PCollection {
```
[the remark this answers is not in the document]
r-jy: I don't think that will work. In sum.go and mean.go (which use dropUnspecifiedPartitionsKVFn), the partition type is embedded in the kv.Pair, and to retrieve it, you have to use pcol.codec.KType, which is an encodedType. What I ended up doing was allowing both the functions to take an EncodedType.

privacy-on-beam/pbeam/aggregations_test.go
```diff
@@ -251,3 +282,149 @@ func TestBoundedSumFloat64FnExtractOutputReturnsNilForSmallPartitions(t *testing
...
+	// In this test, we check  that unspecified partitions
+	// are dropped. This function is used for sum and mean.
+	triples := concatenateTriplesWithFloatValue(
+		makeTripleWithFloatValue(7, 0, 2.0),
+		makeTripleWithFloatValueStartingFromKey(7, 100, 1, 1.3),
+		makeTripleWithFloatValueStartingFromKey(107, 150, 1, 2.5),
+		makeTripleWithFloatValueStartingFromKey(320, 150, 100, 5.5),
+	)
+	// Keep partition 0;
+	// drop partition 1.
```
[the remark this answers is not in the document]
r-jy: This is the same input/output structure used in the mean_test.go function ' TestMeanPerKeyNoNoiseFloatValues'... Since this is a different function I think it makes sense to use the test that I have... I can try to make that clearer in my comment.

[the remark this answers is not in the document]
r-jy: https://github.com/google/differential-privacy/blob/main/privacy-on-beam/pbeam/mean_test.go#L382

privacy-on-beam/pbeam/count.go
```diff
@@ -57,48 +57,66 @@ type CountParams struct {
...
 	// Obtain type information from the underlying PCollection<K,V>.
 	idT, partitionT := beam.ValidateKVType(pcol.col)
-
+	s = s.Scope("pbeam.Count")
 	// Get privacy parameters.
 	spec := pcol.privacySpec
 	epsilon, delta, err := spec.consumeBudget(params.Epsilon, params.Delta)
+	if err != nil {
+		log.Exitf("couldn't consume budget: %v", err)
+	}
```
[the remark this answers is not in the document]
r-jy: So John brought this to my attention last time... and it looks that that part of the code was incorrect previously. Do you still want me to revert my change?

privacy-on-beam/pbeam/count_test.go
```diff
@@ -67,6 +67,40 @@ func TestCountNoNoise(t *testing.T) {
...
+
+	p, s, col, want := ptest.CreateList2(pairs, result)
+	col = beam.ParDo(s, pairToKV, col)
+	partitions := []int{}
+	for i := 9; i < 11; i++ {
+		partitions = append(partitions, i)
+	}
+
+	partitionsCol := CreateList(s, partitions)
+
```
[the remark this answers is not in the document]
r-jy: I'm not sure if flakiness matters here, because partitions are specified. What do you think?

privacy-on-beam/pbeam/count_test.go
r-jy: I am... Since each of 50 users contributes to three partitions, I am making sure that the partitions that are contributed to are those three partitions that I keep. I am testing that the bounds are not wasted on unspecified partitions.

privacy-on-beam/pbeam/count_test.go
r-jy: Same as previous comment about parameters. Should I just add a comment saying that there is no flakiness since the client specifies partitions (is this true?)? I'm just a little confused what sort of comment should go there in this case. What do you think?

privacy-on-beam/pbeam/sum_test.go
r-jy: Same as previous comments... should i just say that epsilon is high, delta is low? I'm not sure that flakiness applies here.

privacy-on-beam/pbeam/sum_test.go
r-jy: Same as previous comment. Testing that bound is not wasted on unspecified partitions.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -305,6 +408,38 @@ func TestSumPerKeyCrossPartitionContributionBoundingInt(t *testing.T) {
...
+		triples = append(triples, makeDummyTripleWithIntValue(50, i)...)
+	}
+	result := []testInt64Metric{
+		{0, 150},
+	}
+	p, s, col, want := ptest.CreateList2(triples, result)
+	col = beam.ParDo(s, extractIDFromTripleWithIntValue, col)
+	partitionsCol := CreateList(s, []int{0, 1, 2})
+
+	epsilon, delta, k, l1Sensitivity := 50.0, 0.01, 25.0, 3.0
```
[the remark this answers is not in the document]
r-jy: Same as previous comments about what I should write there.

privacy-on-beam/pbeam/sum_test.go
r-jy: Same as previous comments. Testing that bound is not wasted on unspecified partitions.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -154,6 +230,33 @@ func TestSumPerKeyNegativeBoundsFloat(t *testing.T) {
...
+	result := []testFloat64Metric{
+		{1, -116.0},
+		{2, -297.0},
+	}
+	p, s, col, want := ptest.CreateList2(triples, result)
+	col = beam.ParDo(s, extractIDFromTripleWithFloatValue, col)
+	partitions := []int{1, 2}
+	partitionsCol := CreateList(s, partitions)
+
+	epsilon, delta, k, l1Sensitivity := 50.0, 1e-200, 25.0, 3.0
```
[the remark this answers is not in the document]
r-jy: Same as previous comments.

privacy-on-beam/pbeam/distinct_id.go
r-jy: You're right, sorry about that!

privacy-on-beam/pbeam/distinct_id.go
```diff
@@ -106,13 +118,28 @@ func DistinctPrivacyID(s beam.Scope, pcol PrivatePCollection, params DistinctPri
...
+		newCountFn(epsilon, delta, maxPartitionsContributed, noiseKind, false),
 		dummyCounts)
 	// Finally, drop thresholded partitions and return the result
 	return beam.ParDo(s, dropThresholdedPartitionsInt64Fn, noisedCounts)
 }
 
+func addSpecifiedPartitionsForDistinctId(s beam.Scope, params DistinctPrivacyIDParams, epsilon, delta float64,
+	maxPartitionsContributed int64, noiseKind noise.Kind, dummyCounts beam.PCollection) beam.PCollection {
+	partitionsCol := params.specifiedPartitions.partitionsCol
+	prepareAddSpecifiedPartitions := beam.ParDo(s, addValuesToSpecifiedPartitionKeysInt64Fn, partitionsCol)
```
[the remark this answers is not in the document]
r-jy: Rename to countsKV right?

privacy-on-beam/pbeam/distinct_id_test.go
```diff
@@ -92,6 +92,50 @@ func TestDistinctPrivacyIDNoNoise(t *testing.T) {
...
+	for i := 4; i < 10000; i++ {
+		result = append(result, testInt64Metric{i, 0})
+	}
+	p, s, col, want := ptest.CreateList2(pairs, result)
+	col = beam.ParDo(s, pairToKV, col)
+
+	partitions := []int{}
+	for i := 0; i < 2; i++ {
+		partitions = append(partitions, i)
+	}
```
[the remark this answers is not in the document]
r-jy: do you mean partitions := []int{0, 1} and then adding the values afterward to the slice?

privacy-on-beam/pbeam/sum.go
```diff
@@ -139,8 +160,13 @@ func SumPerKey(s beam.Scope, pcol PrivatePCollection, params SumParams) beam.PCo
...
 		partialSumPairs,
 		beam.TypeDefinition{Var: beam.XType, T: partitionT})
+
+	if params.specifiedPartitions.partitionsSpecified { // Partitions specified.
+		return addSpecifiedPartitionsForSum(s, epsilon, delta, maxPartitionsContributed,
+			params, noiseKind, vKind, partialSumKV)
+	}
 	sums := beam.CombinePerKey(s,
-		newBoundedSumFn(epsilon, delta, maxPartitionsContributed, params.MinValue, params.MaxValue, noiseKind, vKind),
+		newBoundedSumFn(epsilon, delta, maxPartitionsContributed, params.MinValue, params.MaxValue, noiseKind, vKind, false),
```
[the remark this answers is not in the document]
r-jy: no, it's different... in the function the last parameter is true in this statement the last parameter is false.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -65,6 +63,47 @@ func TestSumPerKeyNoNoiseInt(t *testing.T) {
...
 
+// Checks that SumPerKey with partitions returns a correct answer with int values.
+func TestSumPerKeyWithPartitionsNoNoiseInt(t *testing.T) {
+	triples := concatenateTriplesWithIntValue(
+		makeDummyTripleWithIntValue(7, 0),
+		makeDummyTripleWithIntValue(58, 1),
+		makeDummyTripleWithIntValue(99, 2))
+	for i := 5; i < 10; i++ {
+		v := makeDummyTripleWithIntValue(1, i)
+		triples = append(triples, v...)
```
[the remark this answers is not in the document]
r-jy: Not quite understanding this comment...

privacy-on-beam/pbeam/sum_test.go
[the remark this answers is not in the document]
r-jy: Parameters from previous test are inaccurate... Says there are three partitions but there are only two?

privacy-on-beam/pbeam/aggregations.go
r-jy: This function does not have a return value...Can I use exitf instead?

r-jy: This function does not have a return value...

r-jy: I will use log.Exitf.

privacy-on-beam/pbeam/sum.go
```diff
@@ -103,18 +108,34 @@ func SumPerKey(s beam.Scope, pcol PrivatePCollection, params SumParams) beam.PCo
...
 
 	var noiseKind noise.Kind
+
 	if params.NoiseKind == nil {
 		noiseKind = noise.LaplaceNoise
 		log.Infof("No NoiseKind specified, using Laplace Noise by default.")
 	} else {
 		noiseKind = params.NoiseKind.toNoiseKind()
 	}
+
```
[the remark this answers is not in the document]
r-jy: now that this check depends on noiseKind, I think it makes sense to put it after setting noiseKind.

privacy-on-beam/pbeam/helpers_test.go
```diff
@@ -333,6 +333,27 @@ func approxEqualsKVInt64(s beam.Scope, col1, col2 beam.PCollection, tolerance fl
...
+	}
+	if err := checkValueType(col2, wantV); err != nil {
+		return fmt.Errorf("unexpected value type for col2: %v", err)
+	}
+
+	coGroupToValue := beam.CoGroupByKey(s, col1, col2)
+	diffs := beam.ParDo(s, diffIntFn, coGroupToValue)
+	combinedDiff := beam.Combine(s, combineDiffs, diffs)
+	beam.ParDo0(s, reportDiffs, combinedDiff)
+	return nil
```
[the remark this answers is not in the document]
r-jy: not quite... The type of the partition is an int... It seems non-trivial to convert it to an int64 since the results are PrivatePCollections.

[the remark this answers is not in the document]
r-jy: I use that function to check that partitions are the same.
https://github.com/google/differential-privacy/blob/main/privacy-on-beam/pbeam/helpers_test.go#L201-L205 The fact that partitions are ints is hardcoded in this struct... I tried changing my partitions to an int64 but it didn't really work.

privacy-on-beam/pbeam/aggregations_test.go
```diff
@@ -302,34 +302,30 @@ func TestBoundedSumFloat64FnExtractOutputWithSpecifiedPartitionsDoesNotThreshold
...
-		makePairsWithFixedVStartingFromKey(7+52, 99, 2),
-		makePairsWithFixedVStartingFromKey(7+52+99, 10, 3),
-		makePairsWithFixedVStartingFromKey(7+52+99, 10, 3),
-		makePairsWithFixedVStartingFromKey(7+52+99, 10, 3),
-		makePairsWithFixedVStartingFromKey(7+52+99, 10, 3),
+		makePairsWithFixedV(7, 0),
+		makePairsWithFixedV(52, 1),
+		makePairsWithFixedV(99, 2),
+		makePairsWithFixedV(99, 2),
+		makePairsWithFixedV(99, 2),
```
miracvbasaran, line 318: Do you need to add these 3 times? What are you testing by this? If this is not required, please simply do it once.

privacy-on-beam/pbeam/aggregations_test.go
```diff
@@ -392,23 +389,21 @@ func TestDropUnspecifiedPartitionsInt(t *testing.T) {
 func TestDropUnspecifiedPartitionsFloat(t *testing.T) {
```
miracvbasaran, line 389: I feel like this test should be deleted, we already have test for KV above?

privacy-on-beam/pbeam/count.go
```diff
@@ -125,47 +120,30 @@ func Count(s beam.Scope, pcol PrivatePCollection, params CountParams) beam.PColl
 		newDecodePairInt64Fn(partitionT.Type()),
 		countPairs,
 		beam.TypeDefinition{Var: beam.XType, T: partitionT.Type()})
-
-	if params.specifiedPartitions.partitionsSpecified { // Partitions specified.
-		return addSpecifiedPartitionsForCount(s, epsilon, delta, maxPartitionsContributed,
-			params, noiseKind, countsKV)
-
+	// Add specified partitions, if partitions are specified.
```
miracvbasaran, line 123: Hmm, since we are doing more than adding specified partitions here, maybe we could rename this comment to something like "Add specified partitions and return the aggregation output, if partitions are specified."

(Also for other aggregations)

privacy-on-beam/pbeam/count.go
```diff
@@ -125,47 +120,30 @@ func Count(s beam.Scope, pcol PrivatePCollection, params CountParams) beam.PColl
...
 		return err
 	}
 	err = checks.CheckDeltaStrict("pbeam.Count", delta)
+	if (params.partitionsCol).IsValid() && noiseKind == noise.LaplaceNoise {
+		err = checks.CheckNoDelta("pbeam.Count", delta)
+	} 
 	if err != nil {
 		return err
-	}
+	 }
```
miracvbasaran, line 146: nit: Please remove the extra whitespace here.

privacy-on-beam/pbeam/count.go
```diff
@@ -125,47 +120,30 @@ func Count(s beam.Scope, pcol PrivatePCollection, params CountParams) beam.PColl
...
 }
 
-func checkCountParams(params CountParams, epsilon, delta float64) error {
+func checkCountParams(params CountParams, epsilon, delta float64, noiseKind noise.Kind) error {
 	err := checks.CheckEpsilon("pbeam.Count", epsilon)
 	if err != nil {
 		return err
 	}
 	err = checks.CheckDeltaStrict("pbeam.Count", delta)
+	if (params.partitionsCol).IsValid() && noiseKind == noise.LaplaceNoise {
```
miracvbasaran, line 141: You should make the `CheckDeltaStrict` check in the else statement for this if because otherwise if you have specified partitions and Laplace noise, you would return an error above.

(Also for other aggregations where applicable)

privacy-on-beam/pbeam/count.go
```diff
@@ -175,3 +153,17 @@ func checkCountParams(params CountParams, epsilon, delta float64) error {
 	}
 	return nil
 }
+
+func addSpecifiedPartitionsForCount(s beam.Scope, epsilon, delta float64,
+	maxPartitionsContributed int64, params CountParams, noiseKind noise.Kind, countsKV beam.PCollection) beam.PCollection {
```
miracvbasaran, line 158: Could you please make this single line?

privacy-on-beam/pbeam/count.go
```diff
@@ -175,3 +153,17 @@ func checkCountParams(params CountParams, epsilon, delta float64) error {
...
+	// Turn partitionsCol type PCollection<K> into PCollection<K, int64> by adding
```
miracvbasaran, line 159: "from" instead of "type"

privacy-on-beam/pbeam/count_test.go
```diff
@@ -239,6 +239,57 @@ func TestCountAddsNoise(t *testing.T) {
...
+// Checks that Count with partitions adds noise to its output.
+func TestCountAddsNoiseWithPartitions(t *testing.T) {
+	for _, tc := range []struct {
+		name      string
+		noiseKind NoiseKind
+		// Differential privacy params used.
+		epsilon float64
+		delta   float64
+	}{
+		// Epsilon and delta are not split because partitions are specified. All of it is used for the noise.
```
miracvbasaran, line 251: nit: "All of them are"

privacy-on-beam/pbeam/count_test.go
```diff
@@ -239,6 +239,57 @@ func TestCountAddsNoise(t *testing.T) {
...
+			name:      "Gaussian",
+			noiseKind: GaussianNoise{},
+			epsilon:   1, 
+			delta:     0.005, 
+		},
+		{
+			name:      "Laplace",
+			noiseKind: LaplaceNoise{},
+			epsilon:   0.1, 
+			delta:     0, // It is 0 because partitions are specified (so no thresholding occurs) and we are adding Laplace noise.
```
miracvbasaran, line 262: nit: "we are using Laplace noise."

You could also remove "(so no thresholding occurs)

privacy-on-beam/pbeam/count_test.go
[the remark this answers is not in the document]
miracvbasaran: It still applies because we are adding noise to the aggregation, i.e. the counts. If we don't use a flakiness here, there is a very small chance it might add enough noise to change the output. Does that make sense?

privacy-on-beam/pbeam/count_test.go
```diff
@@ -285,13 +336,16 @@ func TestCountWithPartitionsCrossPartitionContributionBounding(t *testing.T) {
...
 
-	epsilon, delta, k, l1Sensitivity := 50.0, 0.01, 25.0, 3.0
+	// We have 5 partitions. So, to get an overall flakiness of 10⁻²³,
+	// we need to have each partition pass with 1-10⁻²⁵ probability (k=25).
+	epsilon, delta, k, l1Sensitivity := 50.0, 0.0, 25.0, 3.0
 	pcol := MakePrivate(s, col, NewPrivacySpec(epsilon, delta))
-	got := Count(s, pcol, CountParams{MaxPartitionsContributed: 3, MaxValue: 1, NoiseKind: LaplaceNoise{}, specifiedPartitions: partitionsCol})
-	// With a max contribution of 3, all of the data for partitions 0, 1, and 2 should be kept.
+	got := Count(s, pcol, CountParams{MaxPartitionsContributed: 3, MaxValue: 1, NoiseKind: LaplaceNoise{}, partitionsCol: partitionsCol})
+	// With a max contribution of 3, 40% of the data should be
```
miracvbasaran, line 347: Nit: "40% of the data in the specified partitions"

It is a bit ambiguous if the 40% refers to the whole data or the specified partitions data.

privacy-on-beam/pbeam/aggregations_test.go
r-jy: This one's for floats... the previous one was for ints.

privacy-on-beam/pbeam/count.go
r-jy: Good point

privacy-on-beam/pbeam/count.go
r-jy: I'm not sure that's the case. if err from 'checks.CheckDeltaStrict' is not nil and partitions are specified and noiseKind is specified,  then the value of err will change to 'checks.CheckNoDelta'. This all happens before we check if err is nil or not.

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -266,39 +274,44 @@ func (fn *boundedSumInt64Fn) Setup() {
...
+		a.SP.Merge(b.SP)
+	}
 	return a
 }
 
 func (fn *boundedSumInt64Fn) ExtractOutput(a boundedSumAccumInt64) *int64 {
-	result := a.BS.Result()
-	if a.PartitionsSpecified || a.SP.Result() {
+	if a.PartitionsSpecified || a.SP.ShouldKeepPartition(){
+    result := a.BS.Result()
```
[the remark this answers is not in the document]
r-jy: Whoops, sorry

privacy-on-beam/pbeam/count.go
r-jy: Either way, I addressed this comment.

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -450,6 +481,176 @@ func clampNegativePartitionsFloat64Fn(v beam.V, r float64) (beam.V, float64) {
...
+	fn.partitionEnc = beam.NewElementEncoder(fn.PartitionType.T)
+}
+
+func (fn *outputPartitionsNotInTheDataFn) ProcessElement(partitionKey beam.X, value beam.V, partitionsIter func(*PMap) bool, emit func(beam.X, beam.V)) {
+	var partitionBuf bytes.Buffer
+	if err := fn.partitionEnc.Encode(partitionKey, &partitionBuf); err != nil {
+		log.Exitf("pbeam.preparePruneFn.ProcessElement: couldn't encode partition %v: %v", partitionKey, err)
+	}
+	var partitionMap PMap
+	partitionsIter(&partitionMap)
```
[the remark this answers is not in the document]
r-jy: Sure, I mentioned above that this function doesn't have a return value... So I log.Exitf'd instead.

privacy-on-beam/pbeam/sum_test.go
r-jy: Added

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -320,6 +368,60 @@ func TestSumPerKeyAddsNoiseInt(t *testing.T) {
...
+	} {
+		// We have 1 partition. So, to get an overall flakiness of 10⁻²³,
+		// we need to have each partition pass with 1-10⁻²³ probability (k=23).
+		epsilonNoise, deltaNoise := tc.epsilon, tc.delta
+		k := 23.0
+		l0Sensitivity, lInfSensitivity := 1.0, 1.0
+		l1Sensitivity := l0Sensitivity * lInfSensitivity
+		tolerance := complementaryLaplaceTolerance(k, l1Sensitivity, epsilonNoise)
+		if tc.noiseKind == gaussianNoise {
+			deltaNoise = tc.delta
```
miracvbasaran, line 403: We need to remove this line for all such tests. It doesn't do anything.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -187,6 +187,36 @@ func TestBoundedMeanFloat64FnExtractOutputReturnsNilForSmallPartitions(t *testin
...
+	for _, tc := range []struct {
+		desc              string
+		inputSize         int
+		datapointsPerUser int
+	}{
+		{"Empty input", 0, 0},
+		{"Input with 1 user with 1 contribution", 1, 1},
+	} {
+
+		fn := newBoundedMeanFloat64Fn(2*1e100, 1e-23, 1, 1, 0, 10, noise.LaplaceNoise, false)
```
miracvbasaran, line 200: You don't need to use a delta or multiply epsilon by 2, also for other such tests if you are doing it.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -740,27 +904,25 @@ func TestMeanPerKeyWithPartitionsReturnsNonNegativeFloat64(t *testing.T) {
 	for p := 0; p < 200; p++ {
 		partitions = append(partitions, p)
 	}
-	partitionsCol := CreateList(s, partitions)
+	partitionsCol := beam.CreateList(s, partitions)
 
-	// Using a low ε, a high δ, and a high maxValue here to add a
-	// lot of noise while keeping partitions.
-	// ε=0.001. δ=0.999 and l0Sensitivity=1 gives a threshold of =2.
+	// Using a low ε, zero δ, and a high maxValue.
```
miracvbasaran, line 909: nit: "... to add a lot of noise."

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -863,22 +1024,21 @@ func TestMeanPerKeyWithPartitionsCrossPartitionContributionBounding(t *testing.T
...
+	pcol := MakePrivate(s, col, NewPrivacySpec(epsilon, delta))
 	pcol = ParDo(s, tripleWithFloatValueToKV, pcol)
 	got := MeanPerKey(s, pcol, MeanParams{
 		MaxPartitionsContributed:     maxPartitionsContributed,
 		MaxContributionsPerPartition: maxContributionsPerPartition,
 		MinValue:                     lower,
 		MaxValue:                     upper,
 		NoiseKind:                    LaplaceNoise{},
-		specifiedPartitions:          partitionsCol})
+		partitionsCol:          partitionsCol})
```
miracvbasaran, line 1041: Please fix the formatting here (whitespaces)

privacy-on-beam/pbeam/aggregations.go
r-jy: Yep, thank!!

privacy-on-beam/pbeam/mean_test.go
r-jy: Gotcha. Notice how there's no error being returned even though delta is not zero and we are using Laplace noise.. Do you think the check should go in the `BoundedMeanFloat64Fn` rather than in `MeanPerKey`?

privacy-on-beam/pbeam/sum_test.go
r-jy: It was there before.

r-jy: Gotcha

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -330,10 +330,7 @@ func TestMeanPerKeyWithPartitionsAddsNoiseFloat(t *testing.T) {
 		epsilonNoise, deltaNoise := tc.epsilon, tc.delta
 		k := 23.0
 		l0Sensitivity, lInfSensitivity := 1.0, 2.0
```
miracvbasaran, line 332: lInfSensitivity should be 3.0, since lower is 0 and upper is 3.0

And below, when we are calling MeanPerKey, we should use lower and upper instead of hardcoded 0.0 and 2.0 for MinValue and MaxValue.

(Also for the test w/o partitions above, this seems to be an existing bug)

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -536,58 +533,77 @@ func TestMeanPerKeyNoNoiseFloatValues(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("laplaceToleranceForMean: got error %v", err)
-	}
-	if err := approxEqualsKVFloat64(s, got, want, tolerance); err != nil {
-		t.Fatalf("TestMeanPerKeyWithPartitionsNoNoise: %v", err)
-	}
-	if err := ptest.Run(p); err != nil {
-		t.Errorf("TestMeanPerKeyWithPartitionsNoNoise: MeanPerKey(%v) = %v, want %v, error %v", col, got, want, err)
+		want = beam.ParDo(s, float64MetricToKV, want)
+		tolerance, err := laplaceToleranceForMean(25, lower, upper, maxContributionsPerPartition, maxPartitionsContributed, epsilon, 5, exactCount, exactMean)
```
miracvbasaran, line 597: Because laplaceToleranceForMean computes lInfSensitivity based on lower and upper, we don't need to worry about lInfSensitivity ourselves. That's why Mean tests were not flaky.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -536,58 +533,77 @@ func TestMeanPerKeyNoNoiseFloatValues(t *testing.T) {
...
-	})
+		partitions := []int{0, 1}
+		partitionsCol := beam.CreateList(s, partitions)
+
+		// We have ε=50, δ=0 and l0Sensitivity=1. No thresholding is done because partitions are specified.
+		// We have 2 partitions. So, to get an overall flakiness of 10⁻²³,
+		// we can have each partition fail with 10⁻²⁵ probability (k=25).
+		maxContributionsPerPartition := int64(1)
+		maxPartitionsContributed := int64(1)
+		epsilon := 100.0
```
miracvbasaran, line 579: epsilon should be 50 because it's not split.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -1068,55 +1084,74 @@ func TestMeanPerKeyWithPartitionsCrossPartitionContributionBounding(t *testing.T
...
+		exactCount := 250.0
+		exactMean := (100.0 + 2.0*150.0) / exactCount
+
+		// We have ε=50, δ=0 and l0Sensitivity=1.
+		// We do not use thresholding because partitions are specified.
+		// We have 2 partitions. So, to get an overall flakiness of 10⁻²³,
+		// we can have each partition fail with 10⁻²⁵ probability (k=25).
+		maxContributionsPerPartition := int64(1)
+		maxPartitionsContributed := int64(1)
+		epsilon := 10e20
```
miracvbasaran, line 1119: epsilon should be 50 here.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -1068,55 +1084,74 @@ func TestMeanPerKeyWithPartitionsCrossPartitionContributionBounding(t *testing.T
...
+		pcol := MakePrivate(s, col, NewPrivacySpec(epsilon, delta))
+		pcol = ParDo(s, tripleWithIntValueToKV, pcol)
+		partitionsCol := beam.CreateList(s, []int{1, 3})
+		got := MeanPerKey(s, pcol, MeanParams{
+			MaxPartitionsContributed:     maxPartitionsContributed,
+			MaxContributionsPerPartition: maxContributionsPerPartition,
+			MinValue:                     lower,
+			MaxValue:                     upper,
+			NoiseKind:                    LaplaceNoise{},
+			partitionsCol:          partitionsCol})
```
miracvbasaran, line 1143: nit: Please fix the indentation here.

privacy-on-beam/pbeam/mean_test.go
miracvbasaran: Please re-check the epsilon and deltas for all the tests you've written. They are easy to miss as you can see :)

miracvbasaran (changes requested):

privacy-on-beam/pbeam/aggregations.go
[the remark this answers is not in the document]
r-jy: ahhh in this function partitionMap could very well be empty (nil) and it would be working right... because it could be the case that there are no specified partitions already in the data..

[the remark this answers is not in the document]
r-jy: Sorry, I forgot to mention this.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -794,3 +1014,147 @@ func TestFindConvertToFloat64Fn(t *testing.T) {
...
+		exactCount := 250.0
+		exactMean := (100.0 + 2.0*150.0) / exactCount
+
+		// We have ε=50, δ=0 and l0Sensitivity=1.
+		// We do not use thresholding because partitions are specified.
+		// We have 2 partitions. So, to get an overall flakiness of 10⁻²³,
+		// we can have each partition fail with 10⁻²⁵ probability (k=25).
+		maxContributionsPerPartition := int64(1)
+		maxPartitionsContributed := int64(1)
+		epsilon := 50.0
```
r-jy, line 1121: epsilon is not happy here

r-jy: now it is :D

privacy-on-beam/pbeam/aggregations.go
```diff
@@ -664,10 +663,12 @@ func (fn *emitPartitionsNotInTheDataFn) ProcessElement(partitionKey beam.X, valu
...
 		log.Exitf("pbeam.emitPartitionsNotInTheDataFn.ProcessElement: couldn't encode partition %v: %v", partitionKey, err)
 	}
-	var partitionMap PMap
-	partitionsIter(&partitionMap)
+	var partitionsInDataMap PMap
+	partitionsIter(&partitionsInDataMap)
 	encodedPartitionKey := string(partitionBuf.Bytes())
-	if !partitionMap[encodedPartitionKey] {
-		emit(partitionKey, value)
+	// If partitionsInDataMap is nil, partitionMap is empty, so none of the partitions are in the data, which means we need to emit them.
```
miracvbasaran, line 669: nit: `partitionsInDataMap` instead of `partitionMap`

And `which means we need to emit all of them` instead of `which means we need to emit them`

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -1155,3 +1154,77 @@ func TestMeanPerKeyWithPartitionsNoNoiseIntValues(t *testing.T) {
 		}
 	}
 }
+
+// Checks that MeanPerKey with empty specified partitions returns a correct answer for int input values.
```
miracvbasaran, line 1158: nit: Remove `for int input values`

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -1155,3 +1154,77 @@ func TestMeanPerKeyWithPartitionsNoNoiseIntValues(t *testing.T) {
...
+// They should be correctly converted to float64 and then correct result
+// with float statistic should be computed.
```
miracvbasaran, line 1160: Also remove here.

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -1155,3 +1154,77 @@ func TestMeanPerKeyWithPartitionsNoNoiseIntValues(t *testing.T) {
...
+			lower: -10.0,
+			upper: 10.0,
+		},
+	} {
+		triples := concatenateTriplesWithIntValue(
+			makeTripleWithIntValue(7, 0, 2))
+
+		midpoint := tc.lower + (tc.upper-tc.lower)/2.0
+		exactCount := 0.0
+		exactMean := midpoint
```
miracvbasaran, line 1185: Please add an inline comment along the lines of `Mean of 0 elements is midpoint.`

privacy-on-beam/pbeam/mean_test.go
```diff
@@ -1155,3 +1154,77 @@ func TestMeanPerKeyWithPartitionsNoNoiseIntValues(t *testing.T) {
...
+			partitionsCol:                partitionsCol,
+		})
+		want = beam.ParDo(s, float64MetricToKV, want)
+
+		tolerance, err := laplaceToleranceForMean(25, tc.lower, tc.upper, maxContributionsPerPartition, maxPartitionsContributed, epsilon, 0.0, exactCount, exactMean)
+		if err != nil {
+			t.Fatalf("laplaceToleranceForMean: got error %v", err)
+		}
+		if err := approxEqualsKVFloat64(s, got, want, tolerance); err != nil {
+			t.Fatalf("TestMeanPerKeyWithPartitionsNoNoise: %v", err)
```
miracvbasaran, line 1224: Please fix the Test name here and below.

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -114,10 +118,8 @@ func TestSumPerKeyWithPartitionsNoNoiseInt(t *testing.T) {
 
 		partitionsCol := beam.CreateList(s, partitions)
 
-		// We have ε=50, δ=0, and l1Sensitivity=3.
-		// We have 5 partitions. So, to get an overall flakiness of 10⁻²³,
-		// we need to have each partition pass with 1-10⁻²⁵ probability (k=25).
-		epsilon, delta, k, l1Sensitivity := 50.0, 0.0, 25.0, 3.0
+		// We have ε=50, δ=0, and l1Sensitivity=3*lInfSensitivity, to scale with the noise with different MinValues and MaxValues.
```
miracvbasaran, line 121: nit: `to scale the noise`

privacy-on-beam/pbeam/sum_test.go
```diff
@@ -251,7 +257,7 @@ func TestSumPerKeyWithPartitionsNoNoiseFloat(t *testing.T) {
 			triples = append(triples, makeDummyTripleWithFloatValue(1, i) ...)
 		}
 		// Keep partitions 0, 3, and 5.
-		// Drop other partitions up to 100000.
+		// Drop other partitions up to 100.
```
miracvbasaran, line 260: I think you could go even lower, say 10.

miracvbasaran (changes requested):

miracvbasaran (approved):

Issue #18, opened 2020-03-31
Build error in Windows


Dipinka:
Receiving following error after running "bazel build differential_privacy/…"

<path>/differential-privacy/differential_privacy/base/testing/BUILD:38:1: C++ compilation of rule '//differential_privacy/base/testing:status_matchers' failed (Exit 2)
cl : Command line error D8021 : invalid numeric argument '/Wno-sign-compare'
INFO: Elapsed time: 137.683s, Critical Path: 20.36s
INFO: 157 processes: 157 local.
FAILED: Build did NOT complete successfully

dasmdasm:
This is because we add a
`    copts = ["-Wno-sign-compare"],
`
line to every build rule. I guess that compiler flag isn't available on windows. It's not necessary, it just suppresses compiler warnings, so I'll look into making it platform-dependent. In the mean time, if you need a workaround you can remove it from the rules you're trying to build.

aitsam12:
Is it available for windows now? I am still getting this error. 
If not, then how to remove it from rules.

dasmdasm:
Nope, we have not made any progress on this yet. The workaround is to remove the `copts = ["-Wno-sign-compare"],` from the build rules you want to use on Windows.

TedTed:
Do @BenjaminDev's patches solve this error, @Dipinka, or is the workaround still necessary?

BenjaminDev:
@TedTed as far as I'm aware my patches solved the issue. PyDP have the windows build as part of the ci and no one is complaining there. Just a note that PyDP is normally a few commits behind.

@Dipinka if you getting build issues take a look at the[ GitHub Actions for PyDP](https://github.com/OpenMined/PyDP/blob/dev/.github/workflows/pypipublish_windows.yml) as it might shed some light on how we build it.

jasmiem:
[GitHub Actions for PyDP](https://github.com/OpenMined/PyDP/blob/dev/.github/workflows/pypipublish_windows.yml), which is not exist

BenjaminDev:
I've not been working on PyDP in a while and a lot has changed. If `google/differential-privacy` wants to support windows "officially" I'd suggest we add some form of CI that builds for windows - otherwise there will forever be breaks that go unnoticed until PyDP tries to do a release.

[...]

PR #34, opened 2020-08-07
Windows support

BenjaminDev:
This PR addresses #18 (and OpenMined/PyDP#112 of the downstream OpenMined PyDP project)

There are 4 types of issues that blocked the windows build.
1. `std::isnan` on windows does not do an implicit cast to `double` and causes a compiler error: `C2668 'fpclassify': ambiguous call to overloaded function` [See](https://github.com/google/differential-privacy/commit/89dfc542fed2fa9be7d1ec05be56ff2e71789d16)
2. `copts= ["-Wno-sign-compare"]` are not supported on windows. I tried using bazel `features` to turn remove it but either it's not possible or I miss understood how. The solution I used is below. I note this as I'm sure bazel has standard ways of addressing these kinds of use cases but I'm very unfamiliar with bazel. [See](https://github.com/google/differential-privacy/commit/a3c56aaf6664262f17754fc00fc5d3c09bff65ea)
```
    copts = select({
        ":windows": [],
        "//conditions:default": ["-Wno-sign-compare"],
    }),
```
3. The build specifies `c++17` however cl.exe Version 19.26.28806 for x64 does not support designated initializes. A few structs that make use of designated initializes had to be reverted to the previous form. [See](https://github.com/google/differential-privacy/commit/bf24f044e015956540852f50d1b4512fa470da24)
4. The `base/logging.cc` file makes use of a unix only header `unistd.h`. A patch of the form of a `#ifdef` block is needed to conditionally include platform specific alternatives. [See](https://github.com/google/differential-privacy/commit/d998de818689b6b4466ab8c86075498cd948445e)

Aside from these 4 rather small issues (all if which have a proposed single commit fixing them) the code base is fully compatible with windows:grin:!

### DIFF
--- a/cc/base/logging.cc
+++ b/cc/base/logging.cc
@@ -52,7 +52,7 @@ static int clock_gettime(int, struct timespec *spec)      //C-file part
     return 0;
 }
 // END PATCH
-#elif
+#else
 #include <unistd.h>
 #endif
 
--- a/cc/base/logging.cc
+++ b/cc/base/logging.cc
@@ -24,22 +24,22 @@
 #include <sys/types.h>
 #ifdef _WIN32
 // START PATCH
-// HACK: This is not a solution just an attempt at see what is going on and getting PyDP to build on windows.
+// This `if` code block conditionally defines and imports windows system specific code.
+// This block defines the windows alternatives to the functions defined in <unistd.h> 
+// which is included in the `else` section below.
 #include <Windows.h>
-#include <time.h>
+#include <time.h> // for: timespec
 #include <io.h> // for: access 
 #include <direct.h> // for: _mkdir
-
+// By design Windows does not have a `mode` parameter.
 # undef mkdir
-# define mkdir(path, mode) _mkdir(path) // Fix api. Not sure 
+# define mkdir(path, mode) _mkdir(path)  
 
 #ifndef S_ISDIR
 #define S_ISDIR(mode) (((mode) & S_IFMT) == S_IFDIR)
 #endif
 
-#define R_OK 4
 #define W_OK 2
-#define X_OK 0
 #define F_OK 0
 #define CLOCK_REALTIME 0
 
@@ -51,7 +51,6 @@ static int clock_gettime(int, struct timespec *spec)      //C-file part
     spec->tv_nsec = wintime % 10000000i64 * 100;      //nano-seconds
     return 0;
 }
-
 // END PATCH
 #elif
 #include <unistd.h>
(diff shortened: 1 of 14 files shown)


BenjaminDev:
@googlebot I signed it!

cc/base/logging.cc
```diff
@@ -22,7 +22,40 @@
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#ifdef _WIN32
+// START PATCH
+// HACK: This is not a solution just an attempt at see what is going on and getting PyDP to build on windows.
```
dasmdasm, line 27: This comment looks like it's probably out of date. Can you add a comment that says a little more about what this block does?

dasmdasm, line 27: This comment looks like it's probably out of date. Can you add a comment that says a little more about what this block does?

BenjaminDev: > Overall, this looks reasonable.

Awesome!

I have added appropriate comments and noticed I'd introduced 2 unused definitions which have been removed.

The function of this code block is to conditionally define code depending on the system Windows vs Unix. This is necessary as these systems have differing system level api's which are included in `<windows.h>` and `<unistd.h>` as an example. 

**Notes:** 
Moving forward it might be easiest to avoid using these system specific methods as recent c++14/17 standards have `#include <filesystem>` which provides a platform independent interface for call to `IS_DIR`, `mkdir` etc... I'm unsure it covers all syscalls required by this library. I have not gone this route as I'm not very familiar with the core of this code base and potential pitfalls. The proposed approach has the nice property that from a unix perspective the code base is unchanged. 

If there is a desire to move to using the more recent c++ standards to avoid platform specific code I'd be happy to research that and it's implications and report back to this project. However, I see that as a different and potentially more invasive change to the on proposed here.

dasmdasm:
I see cc/base/logging.cc as more of a short-term stopgap than a long-term solution. I don't think it's worth looking into further improving it, as I see replacement in its future.

One further change needed, but otherwise this looks good. I'm done with the internal version, so we can accept this as soon as you make these changes.

cc/base/logging.cc
```diff
@@ -22,7 +22,39 @@
...
+static int clock_gettime(int, struct timespec *spec)      //C-file part
+{
+    __int64 wintime; GetSystemTimeAsFileTime((FILETIME*)&wintime);
+    wintime -= 116444736000000000i64;  //1jan1601 to 1jan1970
+    spec->tv_sec = wintime / 10000000i64;           //seconds
+    spec->tv_nsec = wintime % 10000000i64 * 100;      //nano-seconds
+    return 0;
+}
+// END PATCH
+#elif
```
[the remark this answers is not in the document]
BenjaminDev: I have made the correction.

PR #35, opened 2020-08-07
C++ thresholding primitive

WhimsicottMoon:
Implementation and testing of magic partition selection and Laplace partition selection.


cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+       else if (!epsilon_.has_value()) {
+         return base::InvalidArgumentError("Epsilon has to be set.");
+       }
+       else if (!delta_status.ok()) {
+         return delta_status;
+       }
+       else if (laplace_builder_.has_value()) {
+         std::unique_ptr<NumericalMechanism> mechanism_ =
+           (*(laplace_builder_.value())).SetEpsilon(epsilon_.value())
+           .SetL1Sensitivity(max_partitions_contributed_.value()).Build()
```
dasmdasm, line 261: nit: I think it's a little clearer to use SetL0Sensitivity and SetLinfSensitivity here instead of SetL1Sensitivity, since L0 sensitivity maps directly to max_partitions_contributed_.

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+           .ValueOrDie();
```
dasmdasm, line 262: ValueOrDie is rather dangerous - the code will crash if there was an error. You should return any errors instead (see the ASSIGN_OR_RETURN macro in status_macros.h).

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+
+ protected:
+   double l1_sensitivity_;
+   double diversity_;
+   double threshold_;
+   std::unique_ptr<NumericalMechanism> mechanism_;
+
+};
+
+//We're leaving this be for now
```
dasmdasm, line 315: Please don't leave commented-out code.

cc/algorithms/partition-selection-strategies_test.cc
```diff
@@ -0,0 +1,373 @@
...
+
+using testing::_;
+using testing::Eq;
+using testing::DoubleNear;
+using testing::DoubleEq;
+using testing::MatchesRegex;
+
+//PreaggregationPartitionSelection Tests
+
+TEST(PartitionSelectionStrategiesTest, PreaggPartitionSelectionUnsetEpsilon) {
```
dasmdasm, line 34: Very nit: You can call the test suite "PartitionSelectionTest" to reduce line length.

cc/algorithms/partition-selection-strategies_test.cc
```diff
@@ -0,0 +1,373 @@
...
+
+//We expect the probability of keeping a partition with one user
+//will be approximately delta
+TEST(PartitionSelectionStrategiesTest, PreaggPartitionSelectionOneUser) {
+	PreaggPartitionSelection::Builder test_builder;
+	std::unique_ptr<PartitionSelectionStrategy> build =
+		test_builder.SetEpsilon(0.5).SetDelta(0.02)
+					.SetMaxPartitionsContributed(1).Build().ValueOrDie();
+  	double num_kept = 0.0;
+  	for(int i = 0; i < 1000000; i++) {
```
dasmdasm, line 127: If you're using the same number of samples everywhere make it a file-level constant. If you want to use a unique number of samples for a particular test, make it a local variable in that test (so you don't have to worry about keeping the loop and the probability calculation in sync).

cc/algorithms/partition-selection-strategies_test.cc
```diff
@@ -0,0 +1,373 @@
...
+
+//We expect the probability of keeping a partition
+//with no users will always be zero
+TEST(PartitionSelectionStrategiesTest, PreaggPartitionSelectionNoUsers) {
+	PreaggPartitionSelection::Builder test_builder;
+	std::unique_ptr<PartitionSelectionStrategy> build =
+		test_builder.SetEpsilon(0.5).SetDelta(0.02)
+					.SetMaxPartitionsContributed(1).Build().ValueOrDie();
+  	double num_kept = 0.0;
+  	for(int i = 0; i < 100000; i++) {
```
dasmdasm, line 142: Rather than calculating the percentage kept, just EXPECT_FALSE every time. You can probably reduce the number of samples by a lot there too.

cc/algorithms/partition-selection-strategies_test.cc
```diff
@@ -0,0 +1,373 @@
...
+
+//Values calculated with formula
+TEST(PartitionSelectionStrategiesTest,
+	 PreaggPartitionSelectionNumUsersGreaterThanCrossovers) {
+	PreaggPartitionSelection::Builder test_builder;
+	std::unique_ptr<PartitionSelectionStrategy> build =
+		test_builder.SetEpsilon(0.5).SetDelta(0.02)
+					.SetMaxPartitionsContributed(1).Build().ValueOrDie();
+  	double num_kept = 0.0;
+  	for(int i = 0; i < 10000000; i++) {
```
dasmdasm, line 209: Same thing here - use EXPECT_TRUE rather than calculating the probability.

cc/algorithms/partition-selection-strategies_test.cc
```diff
@@ -15,28 +15,359 @@
...
 using testing::MatchesRegex;
-using testing::Return;
 
-class PartitionSelectionStrategiesTest : public ::testing::Test {};*/
+//PreaggregationPartitionSelection Tests
+
+TEST(PartitionSelectionStrategiesTest, PreaggPartitionSelectionUnsetEpsilon) {
+  	PreaggPartitionSelection::Builder test_builder;
+  	auto failed_build = test_builder.SetDelta(0.1)
+  									.SetMaxPartitionsContributed(2).Build();
```
aforget, line 37: Nit: Is the overabundance of whitespace here and below an artifact of GitHub, or is it actually in the code? If the latter, perhaps it should be shortened to be more sensible?

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+       return epsilon_status;
+     }
+     else if (!delta_status.ok()) {
+       return delta_status;
+     }
+     else if (!max_partitions_contributed_status.ok()) {
+       return max_partitions_contributed_status;
+     }
+     else {
+       std::unique_ptr<PartitionSelectionStrategy> magic =
```
dibakch, line 149: nit: You can directly return and avoid assigning to `magic` (which could otherwise have a better name).

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+    //only keep partition if random number < expected probability of keep
+    return (rand_num <= ProbabilityOfKeep(num_users));
+  }
+
+ protected:
+  double adjusted_epsilon_;
+  double crossover_1_;
+  double crossover_2_;
+
+  double ProbabilityOfKeep(double n) {
```
dibakch, line 211: Could you add a comment what this does and where the calculation comes from?

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+
+};
+
+
+class LaplacePartitionSelection : public PartitionSelectionStrategy {
+ public:
+  //Builder for LaplacePartitionSelection
+  class Builder : public PartitionSelectionStrategy::Builder {
+   public:
+     Builder& SetLaplaceMechanism(LaplaceMechanism::Builder* laplace_builder) {
```
dibakch, line 239: Do you require laplace or could you also use the more generic `NumericalMechanism::Builder`?

cc/algorithms/partition-selection-strategies.h
dibakch: `ValueOrDie` has been recently marked deprecated.  You can use `ASSIGN_OR_RETURN`.

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+       }
+       else if (!epsilon_.has_value()) {
+         return base::InvalidArgumentError("Epsilon has to be set.");
+       }
+       else if (!delta_status.ok()) {
+         return delta_status;
+       }
+       else if (laplace_builder_.has_value()) {
+         std::unique_ptr<NumericalMechanism> mechanism_ =
+           (*(laplace_builder_.value())).SetEpsilon(epsilon_.value())
```
dibakch, line 260: nit: Use `laplace_builder_.value()->SetEpsilon(...)`

( or `(*laplace_builder_)->SetEpsilon(...)` )

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+    return (noised_result > threshold_);
+  }
+
+  double GetL1Sensitivity() { return l1_sensitivity_; }
+
+  double GetDiversity() { return diversity_; }
+
+  double GetThreshold() { return threshold_; }
+
+ protected:
```
dibakch, line 307: Can you make this private?

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+       }
+       else {
+         return base::InvalidArgumentError("The builder has to be set.");
+       }
+     }
+
+   protected:
+     absl::optional<LaplaceMechanism::Builder*> laplace_builder_;
+
+     base::Status EpsilonIsSet() {
```
dibakch, line 277: Is this used somewhere?  I guess it has been inlined and can be removed.

cc/algorithms/partition-selection-strategies.h
WhimsicottMoon: Actually, returning directly will cause an error.  I think it has to do with the unique pointer going out of scope?  Also the name magic is from the name of this method of partition selection, "magic partition selection."  I guess magic does sound kind of vague though, so I renamed it to magic_selection.  Hopefully that fixes it!

cc/algorithms/partition-selection-strategies_test.cc
WhimsicottMoon: I think GitHub may have doubled the size of the indent or something?  SetDelta and SetMaxPartitionsContributed look vertically aligned in my local copy >__<

aforget: I wouldn't worry too much about it. Fry the bigger fish first, and if you don't have time to figure this out, I say don't worry about it.

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+  double GetDelta() { return delta_; }
+
+  double GetAdjustedDelta() { return adjusted_delta_; }
+
+  double GetMaxPartitionsContributed() { return max_partitions_contributed_; }
+
+  virtual bool ShouldKeep(int num_users) = 0;
+
+ protected:
+  double epsilon_;
```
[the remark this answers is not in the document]
WhimsicottMoon: Uh oh... it doesn't work as private.  I think it's because the child classes need to access these variables as well.  So I can't make this change in the base class, but I can do it for the child classes.

cc/algorithms/partition-selection-strategies.h
WhimsicottMoon: I guess I could use a generic NumericalMechanism::Builder?  At least that shouldn't impact the code as it currently is.  But, I think the rationale behind using a Laplace builder instead is that the LaplacePartitionSelection class will calculate a threshold based on the CDF of Laplace, so accepting any type of mechanism that might not be sampling from the Laplace distribution shouldn't happen, which is why the type of the argument is set to a LaplaceMechanism::Builder.

cc/algorithms/partition-selection-strategies.h
```diff
@@ -0,0 +1,345 @@
...
+           .SetL1Sensitivity(max_partitions_contributed_.value()).Build()
+           .ValueOrDie();
+         std::unique_ptr<PartitionSelectionStrategy> laplace =
+           absl::make_unique<LaplacePartitionSelection>(
+             epsilon_.value(), delta_.value(),
+             max_partitions_contributed_.value(), std::move(mechanism_));
+         return laplace;
+       }
+       else {
+         return base::InvalidArgumentError("The builder has to be set.");
```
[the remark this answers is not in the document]
WhimsicottMoon: I fixed this but my solution feels really janky XD

cc/algorithms/partition-selection-strategies_test.cc
WhimsicottMoon: There were two values that were used repeatedly - 10^6 and 10^7.  I made two constants...O__o hope that's okay

cc/algorithms/partition-selection-strategies_test.cc
WhimsicottMoon: I just uh... did one sample for this since the whole test suite is meant to be run many times. Also that made sense with the EXPECT_FALSE to me?  O__o I hope that's sufficient...

cc/algorithms/partition-selection-strategies_test.cc
WhimsicottMoon: Once again, just one sample >__<

cc/algorithms/partition-selection-strategies.h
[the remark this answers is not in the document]
dasmdasm: The style guide's suggestion is to make it private with a protected getter (or setter, as necessary).

cc/algorithms/partition-selection-strategies.h
dasmdasm: There seems to be something missing about our forked version of StatusOr that makes us have to assign to a variable before returning. We have to do this is LaplaceMechanism too.

cc/algorithms/partition-selection-strategies.h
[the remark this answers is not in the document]
dasmdasm: You'll need to update the PR with the new changes before I can comment.

cc/algorithms/partition-selection-strategies_test.cc
dasmdasm: I'd still use more than one sample. If this fails, there's a decent change that it happens only sometimes. We don't want a test that fails sometimes and runs sometimes (that counts as flaky), so multiple samples is still a good idea.

cc/algorithms/partition-selection.h
```diff
@@ -0,0 +1,327 @@
...
+       return max_partitions_contributed_;
+     }    
+
+    private:
+     absl::optional<double> epsilon_;
+     absl::optional<double> delta_;
+     absl::optional<int> max_partitions_contributed_;
+  };
+
+  PartitionSelectionStrategy(double epsilon, double delta,
```
dasmdasm, line 111: Make the constructor protected to force folks to construct it through the builder.

cc/algorithms/partition-selection.h
```diff
@@ -0,0 +1,327 @@
...
+       if (GetEpsilon().value() <= 0) {
+         return base::InvalidArgumentError(
+           absl::StrCat("Epsilon has to be positive but is ",
+                        GetEpsilon().value()));
+       }
+       return base::OkStatus();
+     }
+  };
+
+  PreaggPartitionSelection(double epsilon, double delta, int max_partitions)
```
dasmdasm, line 193: Make this protected, for the same reasons.

cc/algorithms/partition-selection.h
```diff
@@ -0,0 +1,327 @@
...
+                        .SetLInfSensitivity(1).Build());
+       std::unique_ptr<PartitionSelectionStrategy> laplace =
+         absl::make_unique<LaplacePartitionSelection>(
+           GetEpsilon().value(), GetDelta().value(),
+           GetMaxPartitionsContributed().value(), std::move(mechanism_));
+       return laplace;
+     }
+
+   private:
+     absl::optional<std::unique_ptr<LaplaceMechanism::Builder>> laplace_builder_;
```
dasmdasm, line 291: You basically never need an optional<unique_ptr<X>> because the unique_ptr can be nullptr, which already stands for "doesn't have a value".

cc/algorithms/partition-selection_test.cc
```diff
@@ -0,0 +1,362 @@
...
+namespace {
+
+using testing::_;
+using testing::Eq;
+using testing::DoubleNear;
+using testing::DoubleEq;
+using testing::MatchesRegex;
+
+const int num_samples = 10000000;
+const int num_samples_less = 1000000;
```
dasmdasm, line 33: nit: Maybe "small_num_samples"?

dibakch (commented):
Could you add some documentation how a user can use this (maybe as a short Markdown)?

dasmdasm:
> Could you add some documentation how a user can use this (maybe as a short Markdown)?

I think that given the short timeframe here, we can add markdown as a follow-on.

PR #36, opened 2020-08-10
Golang: Noise confidence interval for Laplace noise

kaelhawaty:
- Add Confidence interval for Laplace noise
- Add tests for inverseCDFLaplace and confidence interval interface
  functions

### DIFF
--- a/go/checks/checks_test.go
+++ b/go/checks/checks_test.go
@@ -525,10 +525,10 @@ func TestCheckAlpha(t *testing.T) {
 		{"negative infinity alpha", math.Inf(-1), true},
 		{"infinity alpha", math.Inf(0), true},
 		{"NaN alpha", math.NaN(), true},
-		{"Arbitrary alpha", 0.758464984, false},
+		{"arbitrary alpha", 0.758464984, false},
 	} {
 		if err := CheckAlpha("test", tc.alpha); (err != nil) != tc.wantErr {
-			t.Errorf("checkAlpha: when %s for err got %v, want %t", tc.desc, err, tc.wantErr)
+			t.Errorf("CheckAlpha: when %s for err got %v, want %t", tc.desc, err, tc.wantErr)
 		}
 	}
 }
--- a/go/checks/checks_test.go
+++ b/go/checks/checks_test.go
@@ -513,22 +513,22 @@ func TestCheckMaxPartitionsContributed(t *testing.T) {
 	}
 }
 
-func TestCheckConfidenceLevel(t *testing.T) {
+func TestCheckAlpha(t *testing.T) {
 	for _, tc := range []struct {
-		desc            string
-		ConfidenceLevel float64
-		wantErr         bool
+		desc    string
+		alpha   float64
+		wantErr bool
 	}{
-		{"negative confidenceLevel", -2, true},
-		{"confidenceLevel larger than 1", 10, true},
-		{"positive infinity confidenceLevel", math.Inf(1), true},
-		{"negative infinity confidenceLevel", math.Inf(-1), true},
-		{"infinity confidenceLevel", math.Inf(0), true},
-		{"NaN confidenceLevel", math.NaN(), true},
-		{"random confidenceLevel", 0.758464984, false},
+		{"negative alpha", -2, true},
+		{"alpha larger than 1", 10, true},
+		{"positive infinity alpha", math.Inf(1), true},
+		{"negative infinity alpha", math.Inf(-1), true},
+		{"infinity alpha", math.Inf(0), true},
+		{"NaN alpha", math.NaN(), true},
+		{"Arbitrary alpha", 0.758464984, false},
 	} {
-		if err := CheckConfidenceLevel("test", tc.ConfidenceLevel); (err != nil) != tc.wantErr {
-			t.Errorf("CheckConfidenceLevel: when %s for err got %v, want %t", tc.desc, err, tc.wantErr)
+		if err := CheckAlpha("test", tc.alpha); (err != nil) != tc.wantErr {
+			t.Errorf("checkAlpha: when %s for err got %v, want %t", tc.desc, err, tc.wantErr)
 		}
 	}
 }
--- a/go/noise/laplace_noise.go
+++ b/go/noise/laplace_noise.go
@@ -17,6 +17,7 @@
 package noise
 
 import (
+	"fmt"
 	"math"
 
 	log "github.com/golang/glog"
@@ -164,14 +165,30 @@ func (laplace) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon, 
 	return 1 - math.Pow(1-partitionDelta, float64(l0Sensitivity))
 }
 
+// ConfidenceIntervalInt64 computes a confidence interval that contains int64 noisedValue with probablility
+// greater or equal to the given confidenceLevel with the given noise parameters.
 func (laplace) ConfidenceIntervalInt64(noisedValue, l0Sensitivity, lInfSensitivity int64, epsilon, delta, confidenceLevel float64) (ConfidenceInterval, error) {
-	// TODO: Implement confidence interval computation.
-	return ConfidenceInterval{}, nil
+	if err := checkArgsConfidenceIntervalLaplace("ConfidenceIntervalInt64 (Laplace)", l0Sensitivity, float64(lInfSensitivity), epsilon,
+		delta, confidenceLevel); err != nil {
+		err = fmt.Errorf("ConfidenceIntervalInt64(noisedValue %d, l0sensitivity %d, lInfSensitivity %d, epsilon %f,"+
+			"delta %e, confidenceLevel %f) checks failed with %v", noisedValue, l0Sensitivity, lInfSensitivity, epsilon, delta, confidenceLevel, err)
+		return ConfidenceInterval{}, err
+	}
+	lambda := laplaceLambda(l0Sensitivity, float64(lInfSensitivity), epsilon)
+	return getConfidenceIntervalLaplace(float64(noisedValue), lambda, confidenceLevel).toConfidenceIntervalInt64(), nil
 }
 
+// ConfidenceIntervalFloat64 computes a confidence interval that contains float64 noisedValue with probablility
+// equal to the given confidenceLevel with the given noise parameters.
 func (laplace) ConfidenceIntervalFloat64(noisedValue float64, l0Sensitivity int64, lInfSensitivity, epsilon, delta, confidenceLevel float64) (ConfidenceInterval, error) {
-	// TODO: Implement confidence interval computation.
-	return ConfidenceInterval{}, nil
+	if err := checkArgsConfidenceIntervalLaplace("ConfidenceIntervalFloat64 (Laplace)", l0Sensitivity, lInfSensitivity, epsilon,
+		delta, confidenceLevel); err != nil {
+		err = fmt.Errorf("ConfidenceIntervalFloat64(noisedValue %f, l0sensitivity %d, lInfSensitivity %f, epsilon %f,"+
+			"delta %e, confidenceLevel %f) checks failed with %v", noisedValue, l0Sensitivity, lInfSensitivity, epsilon, delta, confidenceLevel, err)
+		return ConfidenceInterval{}, err
+	}
+	lambda := laplaceLambda(l0Sensitivity, lInfSensitivity, epsilon)
+	return getConfidenceIntervalLaplace(noisedValue, lambda, confidenceLevel), nil
 }
 
 func checkArgsLaplace(label string, l0Sensitivity int64, lInfSensitivity, epsilon, delta float64) error {
@@ -182,6 +199,22 @@ func checkArgsLaplace(label string, l0Sensitivity int64, lInfSensitivity, epsilo
 		return err
 	}
 	if err := checks.CheckEpsilonVeryStrict(label, epsilon); err != nil {
+		return err
+	}
+	return checks.CheckNoDelta(label, delta)
+}
+
+func checkArgsConfidenceIntervalLaplace(label string, l0Sensitivity int64, lInfSensitivity, epsilon, delta, confidenceLevel float64) error {
+	if err := checks.CheckConfidenceLevel(label, confidenceLevel); err != nil {
+		return err
+	}
+	if err := checks.CheckL0Sensitivity(label, l0Sensitivity); err != nil {
+		return err
+	}
+	if err := checks.CheckLInfSensitivity(label, lInfSensitivity); err != nil {
+		return err
+	}
+	if err := checks.CheckEpsilonStrict(label, epsilon); err != nil {
 		return err
 	}
 	return checks.CheckNoDelta(label, delta)
@@ -201,6 +234,32 @@ func addLaplace(x, epsilon, l1Sensitivity float64) float64 {
 func laplaceLambda(l0Sensitivity int64, lInfSensitivity, epsilon float64) float64 {
 	l1Sensitivity := lInfSensitivity * float64(l0Sensitivity)
 	return l1Sensitivity / epsilon
+}
+
+// getConfidenceIntervalLaplace computes a confidence interval that contains float64 noisedValue with probablility
+// equal to the given confidenceLevel with the given lambda.
+func getConfidenceIntervalLaplace(noisedValue float64, lambda, confidenceLevel float64) ConfidenceInterval {
+	// Finding a symmetrical confidence interval around a Laplace of (0, lambda)
+	// by computing alpha, then calculating Z_(1 - alpha/2) using inverseCDFLaplace.
+	alpha := (1 - confidenceLevel)
+	k := 1 - alpha/2
+	Zk := inverseCDFLaplace(0, lambda, k)
+	// The confidence interval will be noisedValue +/- Z_(1-alpha/2).
+	return ConfidenceInterval{noisedValue - Zk, noisedValue + Zk}
+}
+
+// inverseCDFLaplace returns Z_p where if is Y a random variable then Z_p is the value that satisfies the equation:
+// Pr( Y <= Z_p ) = p where mean and lambda are laplace distribution parameters and p is the probablity.
+func inverseCDFLaplace(mean, lambda, p float64) float64 {
+	return (mean - lambda*sign(p-0.5)*math.Log(1-2*math.Abs(p-0.5)))
+}
+
+// sign function helper method.
+func sign(x float64) float64 {
+	if x >= 0 {
+		return 1
+	}
+	return -1
 }
 
 // geometric draws a sample drawn from a geometric distribution with parameter
--- a/go/noise/laplace_noise.go
+++ b/go/noise/laplace_noise.go
@@ -165,8 +165,8 @@ func (laplace) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon, 
 	return 1 - math.Pow(1-partitionDelta, float64(l0Sensitivity))
 }
 
-// ConfidenceIntervalInt64 computes a confidence interval that contains int64 noisedValue with probablility
-// greater or equal to the given confidenceLevel with the given noise parameters.
+// ConfidenceIntervalInt64 computes a confidence interval that contains the true value from which int64 noisedValue
+// was computed with probablility greater or equal to the given confidenceLevel with the given noise parameters.
 func (laplace) ConfidenceIntervalInt64(noisedValue, l0Sensitivity, lInfSensitivity int64, epsilon, delta, confidenceLevel float64) (ConfidenceInterval, error) {
 	if err := checkArgsConfidenceIntervalLaplace("ConfidenceIntervalInt64 (Laplace)", l0Sensitivity, float64(lInfSensitivity), epsilon,
 		delta, confidenceLevel); err != nil {
@@ -178,8 +178,8 @@ func (laplace) ConfidenceIntervalInt64(noisedValue, l0Sensitivity, lInfSensitivi
 	return getConfidenceIntervalLaplace(float64(noisedValue), lambda, confidenceLevel).toConfidenceIntervalInt64(), nil
 }
 
-// ConfidenceIntervalFloat64 computes a confidence interval that contains float64 noisedValue with probablility
-// equal to the given confidenceLevel with the given noise parameters.
+// ConfidenceIntervalFloat64 computes a confidence interval that contains the true value from which float64 noisedValue
+// was computed with probablility equal to the given confidenceLevel with the given noise parameters.
 func (laplace) ConfidenceIntervalFloat64(noisedValue float64, l0Sensitivity int64, lInfSensitivity, epsilon, delta, confidenceLevel float64) (ConfidenceInterval, error) {
 	if err := checkArgsConfidenceIntervalLaplace("ConfidenceIntervalFloat64 (Laplace)", l0Sensitivity, lInfSensitivity, epsilon,
 		delta, confidenceLevel); err != nil {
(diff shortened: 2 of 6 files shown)


go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
 		}
 	}
 }
+func TestInverseCDFLaplace(t *testing.T) {
+	for _, tc := range []struct {
+		desc                     string
+		mean, lambda, prob, want float64
+	}{
+		// 2 high-precision calculated random tests.
```
dkrft, line 180: it is not quite clear what you mean by high-precision (and I am also not sure that it is relevant)

also, I believe you mean "arbitrary" rather than "random"

so i thinkt it suffices to say somethinglike:
"Two arbitrary test cases."

Also, I don't think it is necessary to have a comment at all when the description says essentially the same thing

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+			prob:   0.7875404240919761168041,
+			want:   67.4234254367},
+		{
+			desc:   "High-precision random test",
+			mean:   23,
+			lambda: 2,
+			prob:   0.1479685611330654517049,
+			want:   20.564783456,
+		},
+		// Edge cases where p = 0 or p = 1.(Output should be infinite)
```
dkrft, line 194: since we do not allow confidence levels of 1, it is irrelevant how inverseCDFLaplace behaves on this input. I don't think we should test for this, suggesting that this has some sort of meaning when it has not.

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+			want:   math.Inf(-1),
+		},
+		{
+			desc:   "Positive infinity output when probablity is 1",
+			mean:   1,
+			lambda: 1,
+			prob:   1,
+			want:   math.Inf(1),
+		},
+		// Logical testing (with p = 0.5, return value should be mean regardless of lambda)
```
dkrft, line 209: it is not clear what you mean by "logical"

suggestion:

"For a probablity of 0.5 the result should be the mean regardless of lambda"

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+		{
+			desc:   "%50 confidence level, different lambda",
```
dkrft, line 211: prob is not a confidence level, in particular, prob = 0.5 does not imply 50% confidence.

also it is not clear what "different lambda" is supposed to mean. different from what?

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+
+func TestConfidenceIntervalLaplace(t *testing.T) {
+	// Tests for getConfidenceIntervalLaplace function
+	for _, tc := range []struct {
+		desc                    string
+		noisedValue             float64
+		lambda, confidenceLevel float64
+		want                    ConfidenceInterval
+	}{
+		// 4 Random pre-calculated tests.
```
dkrft, line 257: "arbitrary" instead of "random"

also, "pre-calculated" is clear from context

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+			lambda:          700,
+			confidenceLevel: 0.2,
+			want:            ConfidenceInterval{-90.73838592, 221.6625859},
+		},
+		// Near 0 and 1 confidence levels.
+		{
+			desc:            "Low confidence level",
+			noisedValue:     50,
+			lambda:          10,
+			confidenceLevel: 0.01,
```
dkrft, line 291: i think we should go more extreme in both cases, e.g., 10^-10 and 1-10^-10

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+func TestConfidenceIntervalFloat64(t *testing.T) {
+	for _, tc := range []struct {
+		desc                                             string
+		noisedValue                                      float64
+		l0Sensitivity                                    int64
+		lInfSensitivity, epsilon, confidenceLevel, delta float64
+		want                                             ConfidenceInterval
+		wantErr                                          bool
+	}{
+		{
```
dkrft, line 326: All excpet for one test are argument checks. For the sake of readability and maintainability, I would split this into 2 tests, one for argument checking and one for checking the output (and then maybe add one or two more tests for the second kind of test)

This way, you may also be able to thest the parameters for the int64 and float64 version in the same test, since they should by design have the exact same requirements on their parameters (but ask maria or mirac first if this is good practice)

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+			desc:            "Random test",
```
dkrft, line 327: arbitrary

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,525 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+				got.LowerBound, tc.want.LowerBound, tc.desc)
+		}
+		if !approxEqual(got.UpperBound, tc.want.UpperBound) {
+			t.Errorf("TestConfidenceIntervalFloat64(%f, %d, %f, %f, %f)=%0.10f, want %0.10f, desc %s, UpperBound is not equal",
+				tc.noisedValue, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.confidenceLevel,
+				got.UpperBound, tc.want.UpperBound, tc.desc)
+		}
+	}
+}
+func TestConfidenceIntervalInt64(t *testing.T) {
```
dkrft, line 516: same comments as above

go/noise/laplace_noise.go
```diff
@@ -164,14 +165,30 @@ func (laplace) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
 	return 1 - math.Pow(1-partitionDelta, float64(l0Sensitivity))
 }
 
+// ConfidenceIntervalInt64 computes a confidence interval that contains int64 noisedValue with probablility
```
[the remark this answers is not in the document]
kaelhawaty: Yeah its a typo, should have been "the true value exists with a probablity...". Currently, there was already a merge PR for the interface. I will have to check if I can edit the interface.

go/noise/laplace_noise.go
```diff
@@ -164,14 +165,30 @@ func (laplace) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
...
-	// TODO: Implement confidence interval computation.
-	return ConfidenceInterval{}, nil
+	if err := checkArgsConfidenceIntervalLaplace("ConfidenceIntervalInt64 (Laplace)", l0Sensitivity, float64(lInfSensitivity), epsilon,
+		delta, confidenceLevel); err != nil {
+		err = fmt.Errorf("ConfidenceIntervalInt64(noisedValue %d, l0sensitivity %d, lInfSensitivity %d, epsilon %f,"+
+			"delta %e, confidenceLevel %f) checks failed with %v", noisedValue, l0Sensitivity, lInfSensitivity, epsilon, delta, confidenceLevel, err)
+		return ConfidenceInterval{}, err
+	}
+	lambda := laplaceLambda(l0Sensitivity, float64(lInfSensitivity), epsilon)
+	return getConfidenceIntervalLaplace(float64(noisedValue), lambda, confidenceLevel).toConfidenceIntervalInt64(), nil
```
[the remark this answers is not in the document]
kaelhawaty: Same as above. It might make sense to keep int64 since round does round to integer representable in 64 bits?
https://golang.org/src/math/floor.go?s=1237:1266#L54 not sure though.

go/noise/laplace_noise.go
```diff
@@ -203,6 +236,32 @@ func laplaceLambda(l0Sensitivity int64, lInfSensitivity, epsilon float64) float6
...
+	alpha := (1 - confidenceLevel)
+	k := 1 - alpha/2
+	Zk := inverseCDFLaplace(0, lambda, k)
+	// The confidence interval will be noisedValue +/- Z_(1-alpha/2).
+	return ConfidenceInterval{noisedValue - Zk, noisedValue + Zk}
+}
+
+// inverseCDFLaplace returns Z_p where if is Y a random variable then Z_p is the value that satisfies the equation:
+// Pr( Y <= Z_p ) = p where mean and lambda are laplace distribution parameters and p is the probablity.
+func inverseCDFLaplace(mean, lambda, p float64) float64 {
```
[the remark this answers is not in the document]
kaelhawaty: Calling this "inverseCDFLaplace" without mean as a parameter wouldn't make sense since it would be just a special hlper function to Confidence interval where it returns quarantines for a mean of zero. It made sense to include the mean and make it general in case of needing to use it later on.

go/noise/laplace_noise_test.go
kaelhawaty: It definitely doesn't have a meaning in terms of not allowing confidenceLevels to reach one or zero but it makes sense in terms of checking whether the function converges to infinities when you approach zero or one.

go/noise/laplace_noise_test.go
kaelhawaty: Regardless of the parameter Lambda, you should be getting the mean always with a %50 confidence level. That is what is meant by the test. It should be more clearer, I will adjust it.

go/noise/laplace_noise_test.go
kaelhawaty: I have already discussed this with them before writing the tests. It was better to split the argument checking. "Each test should be concerned only with one function even if they are similar since they might change later on." This was the argument for this approach.

go/noise/laplace_noise.go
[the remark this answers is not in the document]
laviniamnedelea: I also argued about renaming it to roundConfidenceInterval, but the reasoning of not renaming it was, like Karim said, the fact that we "represent" a float64 as an int64.

go/noise/laplace_noise.go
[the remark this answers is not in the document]
kaelhawaty: For now, I added "...that contains the true value from which int64 noisedValue..". Will be changed later on to x and noisedX once we change interface parameters names.

go/noise/laplace_noise_test.go
kaelhawaty: Added, the results for extremely high confidence level were accurate to 8 significant digits.

go/noise/laplace_noise.go
```diff
@@ -203,6 +236,32 @@ func laplaceLambda(l0Sensitivity int64, lInfSensitivity, epsilon float64) float6
...
+	return (mean - lambda*sign(p-0.5)*math.Log(1-2*math.Abs(p-0.5)))
```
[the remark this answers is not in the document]
kaelhawaty: I changed to this but it should be instead:
if p < 1/2
return mean  + lambda * math.Log(2p)
else
return mean - lambda * math.Log(2*(1-p))

because of the already negative sign in the equation.

go/noise/laplace_noise.go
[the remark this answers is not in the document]
kaelhawaty: This was changed with the last commit to https://github.com/google/differential-privacy/blob/1b1dc6639173c0a13613189ec21851604a4c7335/go/noise/noise.go#L69 "toInt64()".

go/noise/laplace_noise.go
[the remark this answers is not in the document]
kaelhawaty: Resolution: Since this is a unexported function "private", It would be easy to add back mean even if we need it later on for other aggregations.

go/noise/laplace_noise_test.go
kaelhawaty: Added low and high probabilities test for convergence.

go/checks/checks_test.go
```diff
@@ -513,22 +513,22 @@ func TestCheckMaxPartitionsContributed(t *testing.T) {
...
-		{"infinity confidenceLevel", math.Inf(0), true},
-		{"NaN confidenceLevel", math.NaN(), true},
-		{"random confidenceLevel", 0.758464984, false},
+		{"negative alpha", -2, true},
+		{"alpha larger than 1", 10, true},
+		{"positive infinity alpha", math.Inf(1), true},
+		{"negative infinity alpha", math.Inf(-1), true},
+		{"infinity alpha", math.Inf(0), true},
+		{"NaN alpha", math.NaN(), true},
+		{"Arbitrary alpha", 0.758464984, false},
```
MashaTelyatnikova, line 528: a - change to lower case

go/checks/checks_test.go
```diff
@@ -513,22 +513,22 @@ func TestCheckMaxPartitionsContributed(t *testing.T) {
...
 	} {
-		if err := CheckConfidenceLevel("test", tc.ConfidenceLevel); (err != nil) != tc.wantErr {
-			t.Errorf("CheckConfidenceLevel: when %s for err got %v, want %t", tc.desc, err, tc.wantErr)
+		if err := CheckAlpha("test", tc.alpha); (err != nil) != tc.wantErr {
+			t.Errorf("checkAlpha: when %s for err got %v, want %t", tc.desc, err, tc.wantErr)
```
MashaTelyatnikova, line 531: CheckAlpha (the name of the function should be exact)

go/noise/laplace_noise.go
```diff
@@ -164,14 +165,30 @@ func (laplace) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
 	return 1 - math.Pow(1-partitionDelta, float64(l0Sensitivity))
 }
 
-func (laplace) computeConfidenceIntervalInt64(noisedValue, l0Sensitivity, lInfSensitivity int64, epsilon, delta, confidenceLevel float64) (ConfidenceInterval, error) {
-	// TODO: Implement confidence interval computation.
-	return ConfidenceInterval{}, nil
+// computeConfidenceIntervalInt64 computes a confidence interval that contains the raw integer value x from which int64 noisedX
+// is computed with a probability greater or equal to 1 - alpha based on the specified laplace noise parameters.
+func (laplace) computeConfidenceIntervalInt64(noisedX, l0Sensitivity, lInfSensitivity int64, epsilon, delta, alpha float64) (ConfidenceInterval, error) {
+	if err := checkArgsConfidenceIntervalLaplace("computeConfidenceIntervalInt64 (Laplace)", l0Sensitivity, float64(lInfSensitivity), epsilon,
```
MashaTelyatnikova, line 171: I think (Laplace) is redundant

go/noise/laplace_noise.go
```diff
@@ -164,14 +165,30 @@ func (laplace) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
...
+		delta, alpha); err != nil {
+		err = fmt.Errorf("computeConfidenceIntervalInt64(noisedX %d, l0sensitivity %d, lInfSensitivity %d, epsilon %f,"+
+			"delta %e, alpha %f) checks failed with %v", noisedX, l0Sensitivity, lInfSensitivity, epsilon, delta, alpha, err)
```
MashaTelyatnikova, line 174: what is %e?

go/noise/laplace_noise.go
```diff
@@ -164,14 +165,30 @@ func (laplace) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
...
-func (laplace) computeConfidenceIntervalFloat64(noisedValue float64, l0Sensitivity int64, lInfSensitivity, epsilon, delta, confidenceLevel float64) (ConfidenceInterval, error) {
-	// TODO: Implement confidence interval computation.
-	return ConfidenceInterval{}, nil
+// computeConfidenceIntervalFloat64 computes a confidence interval that contains the raw value x from which float64
+// noisedX is computed with a probability equal to 1 - alpha based on the specified laplace noise parameters.
+func (laplace) computeConfidenceIntervalFloat64(noisedX float64, l0Sensitivity int64, lInfSensitivity, epsilon, delta, alpha float64) (ConfidenceInterval, error) {
+	if err := checkArgsConfidenceIntervalLaplace("computeConfidenceIntervalFloat64 (Laplace)", l0Sensitivity, lInfSensitivity, epsilon,
+		delta, alpha); err != nil {
+		err = fmt.Errorf("computeConfidenceIntervalFloat64(noisedX %f, l0sensitivity %d, lInfSensitivity %f, epsilon %f,"+
+			"delta %e, alpha %f) checks failed with %v", noisedX, l0Sensitivity, lInfSensitivity, epsilon, delta, alpha, err)
```
MashaTelyatnikova, line 187: what is %e?

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,496 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+			desc:    "Extremely high confidence level",
+			noisedX: 50,
+			lambda:  10,
+			alpha:   7.856382354e-10,
+			want:    ConfidenceInterval{-159.6452468975697118041, 259.6452468975697118041},
+		},
+	} {
+		got := computeConfidenceIntervalLaplace(tc.noisedX, tc.lambda, tc.alpha)
+		if !approxEqual(got.LowerBound, tc.want.LowerBound) {
+			t.Errorf("TestConfidenceIntervalLaplace(%f, %f, %f)=%0.10f, want %0.10f, desc %s, LowerBound is not equal",
```
MashaTelyatnikova, line 281: instead of TestConfidenceIntervalLaplace should be the method name which you test (the same for every other tests)

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,496 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+		},
+	} {
+		got, err := lap.computeConfidenceIntervalInt64(tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity,
+			tc.epsilon, tc.delta, tc.alpha)
+		if (err != nil) != tc.wantErr {
+			t.Errorf("computeConfidenceIntervalInt64: when %v for err got %v, want %t", tc.desc, err, tc.wantErr)
+			continue
+		}
+		if got.LowerBound != tc.want.LowerBound {
+			t.Errorf("TestConfidenceIntervalInt64(%d, %d, %d, %f, %f)=%0.10f, want %0.10f, desc %s, LowerBound is not equal",
```
MashaTelyatnikova, line 656: LowerBound's are not equal?

go/noise/laplace_noise.go
kaelhawaty: Outputs float64 delta in scientific notation, It is used in addNoise(..) functions in Laplace and Gaussian:
https://github.com/google/differential-privacy/blob/1b1dc6639173c0a13613189ec21851604a4c7335/go/noise/laplace_noise.go#L63

I assumed because it is usually extremely small so its not representable using normal "%f"

go/noise/laplace_noise_test.go
kaelhawaty: Yeah I noticed that too but I thought TestComputeConfidenceIntervalLaplace was too long and you don't need to add Compute but I will edit it to follow this rule.

go/noise/laplace_noise.go
kaelhawaty: I followed what was done in the addNoise(..) functions to signify when an error is returned from the noise interface whether it is from Laplace or Gaussian distribution.
https://github.com/google/differential-privacy/blob/1b1dc6639173c0a13613189ec21851604a4c7335/go/noise/laplace_noise.go#L62

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,495 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
 		}
 	}
 }
+func TestInverseCDFLaplace(t *testing.T) {
+	for _, tc := range []struct {
+		desc               string
+		lambda, prob, want float64
```
miracvbasaran, line 178: Could you please have all these fields in their seperate line? I believe that's the idiomatic way of declaring test structs.

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,495 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+		}
+	}
+
+}
+
+func TestComputeConfidenceIntervalLaplace(t *testing.T) {
+	for _, tc := range []struct {
+		desc          string
+		noisedX       float64
+		lambda, alpha float64
```
miracvbasaran, line 232: Also here.

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,495 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+			alpha:           math.NaN(),
+			want:            ConfidenceInterval{},
+			wantErr:         true,
+		},
+	} {
+		got, err := lap.ComputeConfidenceIntervalFloat64(tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity,
+			tc.epsilon, tc.delta, tc.alpha)
+		if (err != nil) != tc.wantErr {
+			t.Errorf("ComputeConfidenceIntervalFloat64: when %v for err got %v, want %t", tc.desc, err, tc.wantErr)
+			continue
```
miracvbasaran, line 476: Does this `continue` do anything? If not, we should remove it.

go/noise/laplace_noise_test.go
```diff
@@ -172,6 +172,495 @@ func TestDeltaForThresholdLaplace(t *testing.T) {
...
+			alpha:           math.NaN(),
+			want:            ConfidenceInterval{},
+			wantErr:         true,
+		},
+	} {
+		got, err := lap.ComputeConfidenceIntervalInt64(tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity,
+			tc.epsilon, tc.delta, tc.alpha)
+		if (err != nil) != tc.wantErr {
+			t.Errorf("ComputeConfidenceIntervalInt64: when %v for err got %v, want %t", tc.desc, err, tc.wantErr)
+			continue
```
miracvbasaran, line 652: Does this `continue` do anything? If not, we should remove it.

miracvbasaran (changes requested):

go/noise/laplace_noise_test.go
kaelhawaty: The idea is if you are getting an error when you are not expecting one, you shouldn't produce an t.Errorf for lower and upper bounds not being equal to the needed confidence interval in the tests logs. Tests logs were more readable this way.

go/noise/laplace_noise.go
```diff
@@ -187,6 +204,22 @@ func checkArgsLaplace(label string, l0Sensitivity int64, lInfSensitivity, epsilo
...
+	}
+	if err := checks.CheckLInfSensitivity(label, lInfSensitivity); err != nil {
+		return err
+	}
+	if err := checks.CheckEpsilonStrict(label, epsilon); err != nil {
+		return err
+	}
+	return checks.CheckNoDelta(label, delta)
+}
+
```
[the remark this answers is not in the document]
kaelhawaty: I thought checks.go is for more general checks where it can be used in multiple files. However, this check is specifically for Laplace distribution and I thought it is better for it to be in this file. Also, this what was done for argument checking in the AddNoise(..) functions
https://github.com/google/differential-privacy/blob/a993d2ae8c5486634559f555647ce169e196bd7a/go/noise/laplace_noise.go#L177

miracvbasaran (approved):

miracvbasaran (approved):

PR #33, opened 2020-08-03
Java: Add confidence interval interface

kaelhawaty:
- Add confidence interval functions to noise interface
- Add value-typed class for confidence Interval with double bounds
- Add confidence level check in DpPreconditions

Fixes:
- Add Infinite condition check for sensitivity parameters in DpPreconditions.Java

PR #41, opened 2020-08-21
Java: Noise confidence interval for Laplace noise

kaelhawaty:
- Add Confidence interval for Laplace noise
- Change interface for confidence interval to alpha instead of confidence level for more accuracy
- Add ConfidenceIntervalLaplaceTest class for tests
- Add different level of strictness to epsilon checks


java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,17 +150,58 @@ public ConfidenceInterval computeConfidenceInterval(
...
       double epsilon,
       @Nullable Double delta,
-      double confidenceLevel) {
-    // TODO: Implement confidence interval computation.
-    throw new UnsupportedOperationException("Not implemented yet.");
+      double alpha) {
+    checkConfidenceParameters(l0Sensitivity, lInfSensitivity, epsilon, delta, alpha);
+    ConfidenceInterval confInt =
+        computeConfidenceInterval(
+            noisedX, Noise.getL1Sensitivity(l0Sensitivity, lInfSensitivity) / epsilon, alpha);
```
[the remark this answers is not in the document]
dkrft: adding a getLambda(...) method makes sense. Feels free to add it and also call it from the other places in the code where lambda is computed. Alternatively, we can file a bug and change this once this is submitted.

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,17 +150,58 @@ public ConfidenceInterval computeConfidenceInterval(
...
   }
 
   private void checkParameters(double l1Sensitivity, double epsilon, @Nullable Double delta) {
-    DpPreconditions.checkEpsilon(epsilon);
+    DpPreconditions.checkEpsilonVeryStrict(epsilon);
     DpPreconditions.checkNoiseDelta(delta, this);
     DpPreconditions.checkL1Sensitivity(l1Sensitivity);
   }
 
+  private void checkConfidenceParameters(
```
dkrft, line 168: nit: make sure that naming is consistent with GaussianNoise

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,17 +150,58 @@ public ConfidenceInterval computeConfidenceInterval(
...
+      int l0Sensitivity,
+      double lInfSensitivity,
+      double epsilon,
+      @Nullable Double delta,
+      double alpha) {
+    DpPreconditions.checkEpsilonStrict(epsilon);
```
dkrft, line 174: I know that epsilon does not have to be very strict for this to work, but since all sorts of stuff may happen for super small epsilon (overflows, arithmetic imprecision, etc.) and we can't compute noise for these parameters anyway (not that there would be a use case for it) we can make our lives easier and just use checkEpsilonVeryStrict(epsilon). In particular, you can reuse checkParameters here.

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,17 +150,58 @@ public ConfidenceInterval computeConfidenceInterval(
...
+    DpPreconditions.checkNoiseDelta(delta, this);
+    DpPreconditions.checkSensitivities(l0Sensitivity, lInfSensitivity);
+    DpPreconditions.checkAlpha(alpha);
+  }
+
+  /**
```
dkrft, line 180: nit: ideally, the style of this doc is consistent with that of Gaussian noise. I made a suggestion in the Gaussian noise version that is a bit more succinct and consistent with the addNoise doc. 

other nit: "{@code lambda}" rather than "lambda"

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,17 +150,58 @@ public ConfidenceInterval computeConfidenceInterval(
...
+   * Computes a confidence interval that contains the raw value {@code x} passed to {@link
+   * #addNoise(double, int, double, double, Double)} with a probability equal to {@code 1 - alpha}
+   * based on the specified {@code noisedX} and lambda.
+   */
+  private ConfidenceInterval computeConfidenceInterval(
+      double noisedX, double lambda, double alpha) {
+    // Finding a symmetrical confidence interval around a Laplace of (0, lambda)
```
dkrft, line 187: As I mentioned in the Gaussian noise implementation, there is no need for a comment here.

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,17 +150,58 @@ public ConfidenceInterval computeConfidenceInterval(
...
+    // by calculating Z_(alpha/2) using inverseCdfLaplace which will return a
+    // negative value by symmetry to gain more accuracy for extremely small alpha.
+    double z = inverseCdfLaplace(lambda, alpha / 2);
+    return ConfidenceInterval.create(noisedX + z, noisedX - z);
+  }
+
+  /**
+   * Computes the quantile z satisfying Pr[Y <= z] = p for a random variable Y that is Laplace
```
dkrft, line 195: the doc is good but please try to be consistent with the Gaussian implementation.

some nits: 
- "percentile" rather than "quantile"
- "{@code ...}" around "p" and "lambda"
- "and a mean of 0." rather than "where mean is zero."

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,17 +150,58 @@ public ConfidenceInterval computeConfidenceInterval(
...
+   * distributed with the specified lambda where mean is zero.
+   */
+  private double inverseCdfLaplace(double lambda, double p) {
```
dkrft, line 198: as I mentioned for Gaussian noise, this can be static and consider naming the method as an action, e.g., "computeLaplacePercentile(...)"

java/tests/com/google/privacy/differentialprivacy/ConfidenceIntervalLaplaceTest.java
```diff
@@ -0,0 +1,276 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import javax.annotation.Nullable;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.Collection;
+
+@RunWith(Enclosed.class)
+public class ConfidenceIntervalLaplaceTest {
```
dkrft, line 18: Same comments as for the Gaussian tests apply

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
dkrft, line 157: This step implicitly casts a long to a double, which can be a problem for large long values. See my comment on the Gaussian noise version of this for a potential fix. Also, make sure the implement the fix consistently between Laplace and Gaussian noise. It may make sense to add some of the math functionality to the SecureMathUtils so that you can share them without the need to re-implement.

dkrft (changes requested):

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
kaelhawaty: Yeah I agree but this was requested in the Java interface PR and was implemented in Go and merged.
https://github.com/google/differential-privacy/pull/33#discussion_r467125967
Arithmetic imprecision might come from lambda computation and would return infinite bounds in case of an overflow which makes sense if you are using values on that scale. Alternatively, we can change this here and make a PR to Go to change it.

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
kaelhawaty: @laviniamnedelea I think we can use this documentation with those changes since it is what we used in Go too.

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
kaelhawaty: Most of the functions that are in this file could also be static, so I think having this non-static to follow the trend is better because it will be the only static function in the file and since its private, I don't think there is enough reason to change it to static. Is there is a reason to why it should be static?

For function name, method as an action is better for Java. I will change it. Thanks!

java/main/com/google/privacy/differentialprivacy/LaplaceNoise.java
```diff
@@ -149,9 +152,14 @@ public ConfidenceInterval computeConfidenceInterval(
...
       double epsilon,
       @Nullable Double delta,
-      double confidenceLevel) {
-    // TODO: Implement confidence interval computation.
-    throw new UnsupportedOperationException("Not implemented yet.");
+      double alpha) {
+    checkConfidenceIntervalParameters(l0Sensitivity, lInfSensitivity, epsilon, delta, alpha);
+    ConfidenceInterval confInt =
+        computeConfidenceInterval(
+            0, Noise.getL1Sensitivity(l0Sensitivity, lInfSensitivity) / epsilon, alpha);
```
MashaTelyatnikova, line 159: I would add variable lambda here, would be easier to read.

Also I would rename it to either  interval[AroundZero] or confidenceInterval[AroundZero]. WDYT?

java/main/com/google/privacy/differentialprivacy/SecureNoiseMath.java
```diff
@@ -85,4 +85,18 @@ public static double roundToMultipleOfPowerOfTwo(double x, double granularity) {
...
+  /** Computes the closest double value that is larger than or equal to the provided long value. */
+  public static double nextLargerDouble(long n) {
+    double result = n;
+    long dif = (long) result - n;
+    return dif >= 0 ? result : Math.nextUp(result);
+  }
+
+  /** Computes the closest double value that is smaller than or equal to the provided long value. */
+  public static double nextSmallerDouble(long n) {
+    double result = n;
```
MashaTelyatnikova, line 98: can you please add an example in the comments? input and output and why this works?

java/tests/com/google/privacy/differentialprivacy/BUILD
```diff
@@ -14,14 +14,14 @@
 # limitations under the License.
 
 package(default_visibility = [
-    "//visibility:public"
+    "//visibility:public",
```
MashaTelyatnikova, line 17: remove this change please

java/tests/com/google/privacy/differentialprivacy/BUILD
```diff
@@ -14,14 +14,14 @@
...
 ])
 
 java_plugin(
     name = "autovalue_plugin",
     processor_class = "com.google.auto.value.processor.AutoValueProcessor",
     deps = [
-        "@maven//:com_google_auto_value_auto_value"
+        "@maven//:com_google_auto_value_auto_value",
```
MashaTelyatnikova, line 24: remove this change please

java/tests/com/google/privacy/differentialprivacy/BUILD
```diff
@@ -44,13 +44,15 @@ java_test(
...
     deps = [
-               ":autovalue",
-               "//main/com/google/privacy/differentialprivacy",
-               "@com_google_differential_privacy//proto:summary_java_proto",
-               "@maven//:com_google_guava_guava",
-               "@maven//:com_google_truth_truth",
-               "@maven//:junit_junit",
-               "@maven//:org_mockito_mockito_core",
-               "@maven//:com_google_protobuf_protobuf_java",
+        ":autovalue",
```
MashaTelyatnikova, line 47: why indentation is different from what it was before?

java/tests/com/google/privacy/differentialprivacy/LaplaceNoiseConfidenceIntervalTest.java
```diff
@@ -0,0 +1,649 @@
...
+  private static final double DEFAULT_EPSILON = 0.1;
+  private static final Double DEFAULT_DELTA = null;
+  private static final double DEFAULT_ALPHA = 0.1;
+
+  private boolean approxEqual(double a, double b) {
+    double mxMagnitude = Math.max(Math.abs(a), Math.abs(b));
+    return Math.abs(a - b) <= TOLERANCE * mxMagnitude;
+  }
+
+  private void isEqual(ConfidenceInterval a, ConfidenceInterval b) {
```
MashaTelyatnikova, line 26: you can use approxEqual everywhere I guess

java/tests/com/google/privacy/differentialprivacy/LaplaceNoiseConfidenceIntervalTest.java
```diff
@@ -0,0 +1,649 @@
...
+        .that(approxEqual(a.lowerBound(), b.lowerBound()))
+        .isTrue();
+    assertWithMessage("Upper bounds are not equal.")
+        .that(approxEqual(a.upperBound(), b.upperBound()))
+        .isTrue();
+  }
+
+  @Test
+  public void computeConfidenceInterval_double_arbitraryTest() {
+    ConfidenceInterval confInt =
```
MashaTelyatnikova, line 48: rename to actual and everywhere else

java/tests/com/google/privacy/differentialprivacy/LaplaceNoiseConfidenceIntervalTest.java
```diff
@@ -0,0 +1,649 @@
...
+        NOISE.computeConfidenceInterval(
+            /* noisedX */ 50,
+            DEFAULT_L_0_SENSITIVITY,
+            DEFAULT_L_INF_SENSITIVITY,
+            DEFAULT_EPSILON,
+            DEFAULT_DELTA,
+            /* alpha */ 7.856382354e-10);
+    ConfidenceInterval expected = ConfidenceInterval.create(-159.645246897, 259.645246897);
+    isEqual(confInt, expected);
```
MashaTelyatnikova, line 86: I would rename to "verifyApproxEqual"

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/SecureNoiseMath.java
MashaTelyatnikova: the same in the above method

java/tests/com/google/privacy/differentialprivacy/LaplaceNoiseConfidenceIntervalTest.java
```diff
@@ -0,0 +1,649 @@
...
+            DEFAULT_L_INF_SENSITIVITY,
+            /* epsilon */ 0.3,
+            DEFAULT_DELTA,
+            /* alpha */ 0.05);
+    ConfidenceInterval expected = ConfidenceInterval.create(3.014225755, 22.98577425);
+    isEqual(confInt, expected);
+  }
+
+  @Test
+  public void computeConfidenceInterval_double_lowConfidenceLevel() {
```
MashaTelyatnikova, line 61: forDouble
forLong

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/SecureNoiseMath.java
kaelhawaty: Added explanation to the documentation and an example in comments.

kaelhawaty:
Rebased to pull new format changes to the BUILD file.

MashaTelyatnikova (approved):

PR #46, opened 2020-08-26
Java: Confidence interval for BoundedSum aggregation.

laviniamnedelea:


java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +125,38 @@ public double computeResult() {
...
+            params
+                    .noise()
+                    .computeConfidenceInterval(
+                            noisedSum,
+                            params.maxPartitionsContributed(),
+                            (double) params.maxContributionsPerPartition(),
+                            params.epsilon(),
+                            params.delta(),
+                            alpha);
+    if ((params.lower() < 0) && (params.upper() <= 0)) {
```
MashaTelyatnikova, line 152: can we do it with less amounts of Ifs?
 like in Go

	// If lower and upper bounds are non-negative, trim the negative part of the interval.
	if bs.lower >= 0 {
		confInt.LowerBound, confInt.UpperBound = math.Max(0, confInt.LowerBound), math.Max(0, confInt.UpperBound)
	}
	// Similarly if lower and upper bounds are non-positive, trim the positive part of the interval.
	if bs.upper <= 0 {
		confInt.LowerBound, confInt.UpperBound = math.Min(0, confInt.LowerBound), math.Min(0, confInt.UpperBound)
	}

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +680,215 @@ private static void testForBias(
     // equal to the raw count.
     assertThat(stats.mean()).isWithin(sampleTolerance).of(rawEntry);
   }
-}
+
+  @Test
+  public void computeConfidenceInterval_negativeSumBounds() {
```
MashaTelyatnikova, line 685: I thinks all comments to the Count conf intervals tests apply here as well

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +680,215 @@ private static void testForBias(
...
+    verify(noise)
+            .computeConfidenceInterval(
+                    eq(0.0),
+                    eq(1),
+                    eq(1.0),
+                    eq(EPSILON),
+                    eq(DELTA),
+                    eq(alpha));
+  }
+  }
```
MashaTelyatnikova, line 894: add new line at the end of the file

MashaTelyatnikova (changes requested):

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,213 @@ private static void testForBias(
...
```
MashaTelyatnikova, line 890: looks like something is wrong with the formatting, please re-format the whole file

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +125,38 @@ public double computeResult() {
...
-        .addNoise(sum, getL0Sensitivity(), lInfSensitivity, params.epsilon(), params.delta());
+            getLInfSensitivity(params.lower(), params.upper(), params.maxContributionsPerPartition());
+    noisedSum =
+            params
+                  .noise()
+                  .addNoise(sum, getL0Sensitivity(), lInfSensitivity, params.epsilon(), params.delta());
+    return noisedSum;
+  }
+
+  /** ComputeConfidenceInterval computes a confidence interval that contains the true {@link BoundedSum} with
```
MashaTelyatnikova, line 136: formatting seems not right, please re-format the file

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,216 @@ private static void testForBias(
     // equal to the raw count.
     assertThat(stats.mean()).isWithin(sampleTolerance).of(rawEntry);
   }
+
+  @Test
+  public void computeConfidenceInterval_negativeSumBounds() {
```
MashaTelyatnikova, line 683: add the result

"clampsToZero"

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,216 @@ private static void testForBias(
...
+            .thenReturn(ConfidenceInterval.create(-5, 3));
+    sum.computeResult();
+
+    // The result interval = (-5, 3), but the bounds are negative, therefore it should be clamped to (-5, 0).
+    assertThat(sum.computeConfidenceInterval(0.152145599))
+            .isEqualTo(ConfidenceInterval.create(-5, 0));
+  }
+
+  @Test
+  public void computeConfidenceInterval_positiveSumBounds() {
```
MashaTelyatnikova, line 704: add theexpected  result

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,216 @@ private static void testForBias(
...
+    sum.computeResult();
+
+    // The result interval = (Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY), but the bounds are positive,
+    // therefore it should be clamped to (0, Double.POSITIVE_INFINITY).
+    assertThat(sum.computeConfidenceInterval(0.152145599))
+            .isEqualTo(ConfidenceInterval.create(0, Double.POSITIVE_INFINITY));
+  }
+
+  @Test
+  public void computeConfidenceInterval__forGaussianNoise() {
```
MashaTelyatnikova, line 768: extra _ before "forGaussian"

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +125,38 @@ public double computeResult() {
...
+            params
+                    .noise()
+                    .computeConfidenceInterval(
+                            noisedSum,
+                            params.maxPartitionsContributed(),
+                            (double) params.maxContributionsPerPartition(),
+                            params.epsilon(),
+                            params.delta(),
+                            alpha);
+    if (params.lower() >= 0) {
```
MashaTelyatnikova, line 153: can you please add tests for both ifs?

Like for example, lower >= 0 and you should do clamping
or if lower <. 0 and negative bounds you don't do clamping

the same for upper

Consider not only if condition (min and max are also basically if-condition, so consider all cases)

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,216 @@ private static void testForBias(
...
+    BoundedSum.builder()
+            .epsilon(EPSILON)
+            .delta(DELTA)
+            .noise(noise)
+            .maxPartitionsContributed(1)
+            .lower(1)
+            .upper(5)
+            .build();
+    sum.addEntry(10);
+    Exception exception = assertThrows(IllegalStateException.class, () -> {
```
MashaTelyatnikova, line 831: make Exception IllegalStateException (in other cases as well)

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,216 @@ private static void testForBias(
...
+
+  @Test
+  public void computeConfidenceInterval_whenComputeResultWasNotCalled_throwsException_forLong() {
+    // Mock the noise mechanism. Since noise is not Laplace, nor Gaussian, delta will be passed
+    // as a value instead of null, in order to pass the checks.
+    when(noise.computeConfidenceInterval(anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
+            .thenAnswer(invocation -> new LaplaceNoise().computeConfidenceInterval(
+                    (Long) invocation.getArguments()[0], (Integer) invocation.getArguments()[1], (Long) invocation.getArguments()[2],
+                    (Double) invocation.getArguments()[3], null, (Double) invocation.getArguments()[5]));
+    BoundedSum.builder()
```
MashaTelyatnikova, line 822: can't we use default sum?

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,216 @@ private static void testForBias(
...
+
+  @Test
+  public void computeConfidenceInterval_whenComputeResultWasNotCalled_throwsException_forDouble() {
+    // Mock the noise mechanism. Since noise is not Laplace, nor Gaussian, delta will be passed
+    // as a value instead of null, in order to pass the checks.
+    when(noise.computeConfidenceInterval(anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
+            .thenAnswer(invocation -> new LaplaceNoise().computeConfidenceInterval(
+                    (Double) invocation.getArguments()[0], (Integer) invocation.getArguments()[1], (Double) invocation.getArguments()[2],
+                    (Double) invocation.getArguments()[3], null, (Double) invocation.getArguments()[5]));
+    BoundedSum.builder()
```
MashaTelyatnikova, line 847: default sum?

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +678,216 @@ private static void testForBias(
...
+
+  @Test
+  public void computeConfidenceIntervals_defaultParameters_callsNoiseCorrectly() {
+    double alpha = 0.1524;
+    when(noise.computeConfidenceInterval(anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
+            .thenAnswer(invocation -> new GaussianNoise().computeConfidenceInterval(
+                    (Double) invocation.getArguments()[0], (Integer) invocation.getArguments()[1], (Double) invocation.getArguments()[2],
+                    (Double) invocation.getArguments()[3], (Double) invocation.getArguments()[4], (Double) invocation.getArguments()[5]));
+    sum =
+            BoundedSum.builder()
```
MashaTelyatnikova, line 872: maybe default sum?

MashaTelyatnikova (changes requested):

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
laviniamnedelea: Default sum has -+Double.MAX_VALUE lower and upper bounds. For consistency with other tests i think it would be better to change the lower and upper bounds by building sum this way. Should i change it, WDYT?

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,299 @@ private static void testForBias(
...
   }
+
+  @Test
+  public void computeConfidenceInterval_negativeSumBounds_noClamping() {
+    sum =
+            BoundedSum.builder()
+                    .epsilon(EPSILON)
+                    .delta(DELTA)
+                    .noise(noise)
+                    .maxPartitionsContributed(12)
```
MashaTelyatnikova, line 691: why 12? do we really need this in this test? From the first glance default value (1) should be ok here.

MashaTelyatnikova: I would add method

BoundedSum.Params.Builder getBoundedSumBuilderWithFields() {
    return BoundedSum.builder()
                    .epsilon(EPSILON)
                    .delta(DELTA)
                    .noise(noise)
                    .maxPartitionsContributed(1)
                    .maxContributionsPerPartition(1)
                    .lower(-Double.MAX_VALUE)
                   .upper(Double.MAX_VALUE);
}


and I would use it everywhere where I just need to change few parameters (which differ from default)


in the setup method

sum = getBoundedSumBuilderWithFields().build();


in this test

sum = getBoundedSumBuilderWithFields().lower(-8).upper(-2).build();

so we clearly see that in order to check the need behavior in this test we need those 2 params (and others don't matter)
}

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,299 @@ private static void testForBias(
...
+                    .lower(-8)
+                    .upper(-2)
+                    .build();
+    when(noise.computeConfidenceInterval(
+            anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
+            .thenReturn(ConfidenceInterval.create(-5, -3));
+    sum.computeResult();
+
+    // The result interval = (-5, -3), but the bounds are negative, therefore they shouldn't be clamped.
```
MashaTelyatnikova, line 700: remove such comments like in Count

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,299 @@ private static void testForBias(
...
+    sum.computeResult();
+
+    assertThat(sum.computeConfidenceInterval(0.5)).isEqualTo(ConfidenceInterval.create(3.068528194400547, 16.931471805599454));
+  }
+
+  @Test
+  public void computeConfidenceInterval_computeResultWasNotCalled_throwsException_forLong() {
+    // Mock the noise mechanism. Since noise is not Laplace, nor Gaussian, delta will be passed
+    // as a value instead of null, in order to pass the checks.
+    when(noise.computeConfidenceInterval(anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
```
MashaTelyatnikova, line 904: we don't probably need this and can use default mocked noise (like we did in Count)

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,299 @@ private static void testForBias(
...
+    BoundedSum.builder()
+            .epsilon(EPSILON)
+            .delta(DELTA)
+            .noise(noise)
+            .maxPartitionsContributed(1)
+            .lower(1)
+            .upper(5)
+            .build();
+    IllegalStateException exception = assertThrows(IllegalStateException.class, () -> {
+      sum.computeConfidenceInterval(0.1554684);
```
MashaTelyatnikova, line 942: what is the reason to use such alpha? why not default one? the same for the next test

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,299 @@ private static void testForBias(
...
+    });
+    assertThat(exception)
+            .hasMessageThat()
+            .startsWith("computeResult must be called before calling computeConfidenceInterval.");
+  }
+
+  @Test
+  public void computeConfidenceIntervals_defaultParameters_callsNoiseCorrectly() {
+    double alpha = 0.1524;
+    when(noise.computeConfidenceInterval(anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
```
MashaTelyatnikova, line 952: we don't probably need add when and we can use default mocked noise (like in Count)

MashaTelyatnikova (changes requested):

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,227 @@ private static void testForBias(
...
+    sum = getBoundedSumBuilderWithFields().lower(1.0).upper(10.0).build();
+    sum.addEntry(1);
+    sum.computeResult();
+
+    assertThat(sum.computeConfidenceInterval(ALPHA))
+            .isEqualTo(ConfidenceInterval.create(0.0, 154.0827096308998));
+  }
+
+  @Test
+  public void computeConfidenceInterval_computeResultWasNotCalled_throwsException_forLong() {
```
MashaTelyatnikova, line 828: forLong and forDouble should be after "computeResultWasnotCalled"

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +126,38 @@ public double computeResult() {
...
-        .addNoise(sum, getL0Sensitivity(), lInfSensitivity, params.epsilon(), params.delta());
+            getLInfSensitivity(params.lower(), params.upper(), params.maxContributionsPerPartition());
+    noisedSum =
+            params
+                  .noise()
+                  .addNoise(sum, getL0Sensitivity(), lInfSensitivity, params.epsilon(), params.delta());
+    return noisedSum;
+  }
+
+  /** ComputeConfidenceInterval computes a confidence interval that contains the true {@link BoundedSum} with
```
MashaTelyatnikova, line 137: something is wrong with formatting, can you do the same (as we did in Count)?

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,227 @@ private static void testForBias(
...
+            anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
+            .thenReturn(ConfidenceInterval.create(-3.0, -1.0));
+    sum.computeResult();
+
+    assertThat(sum.computeConfidenceInterval(ALPHA))
+            .isEqualTo(ConfidenceInterval.create(0.0, 0.0));
+  }
+
+  @Test
+  public void computeConfidenceInterval_differentSumBoundsSigns() {
```
MashaTelyatnikova, line 757: add a expected result to the name (below as well)

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +680,227 @@ private static void testForBias(
...
+      sum.computeConfidenceInterval(ALPHA);
+    });
+    assertThat(exception)
+            .hasMessageThat()
+            .startsWith("computeResult must be called before calling computeConfidenceInterval.");
+  }
+
+  @Test
+  public void computeConfidenceIntervals_defaultParameters_callsNoiseCorrectly() {
+    when(noise.computeConfidenceInterval(
```
MashaTelyatnikova, line 881: like with Count we probably don't need to do mocking again? we can use default

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +125,38 @@ public double computeResult() {
...
+                    .noise()
+                    .computeConfidenceInterval(
+                            noisedSum,
+                            params.maxPartitionsContributed(),
+                            (double) params.maxContributionsPerPartition(),
+                            params.epsilon(),
+                            params.delta(),
+                            alpha);
+    if ((params.lower() < 0) && (params.upper() <= 0)) {
+      if (confInt.lowerBound() > 0) confInt = ConfidenceInterval.create(0, confInt.upperBound());
```
dkrft: this is much more compact

return ConfidenceInterval.create(Math.min(0.0, confInt.upperBound()), Math.min(0.0, confInt.upperBound()));

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +125,38 @@ public double computeResult() {
...
+      if (confInt.upperBound() > 0) confInt = ConfidenceInterval.create(confInt.lowerBound(), 0);
+    } else if ((params.lower() >= 0) && (params.upper() > 0)) {
+      if (confInt.lowerBound() < 0) confInt = ConfidenceInterval.create(0, confInt.upperBound());
```
dkrft: more compact

return ConfidenceInterval.create(Math.max(0.0, confInt.upperBound()), Math.max(0.0, confInt.upperBound()));

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
dkrft: if (params.upper() <= 0.0) {

the lower bound is implicitly less than 0 in this case

also, it is good style to write mark double falues as such, e.g. by writing zero as 0.0

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +125,38 @@ public double computeResult() {
...
+                            noisedSum,
+                            params.maxPartitionsContributed(),
+                            (double) params.maxContributionsPerPartition(),
+                            params.epsilon(),
+                            params.delta(),
+                            alpha);
+    if ((params.lower() < 0) && (params.upper() <= 0)) {
+      if (confInt.lowerBound() > 0) confInt = ConfidenceInterval.create(0, confInt.upperBound());
+      if (confInt.upperBound() > 0) confInt = ConfidenceInterval.create(confInt.lowerBound(), 0);
+    } else if ((params.lower() >= 0) && (params.upper() > 0)) {
```
dkrft: else if (params.lower() >= 0.0) {

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -124,10 +125,38 @@ public double computeResult() {
...
+      if (confInt.lowerBound() < 0) confInt = ConfidenceInterval.create(0, confInt.upperBound());
+      if (confInt.upperBound() < 0) confInt = ConfidenceInterval.create(confInt.lowerBound(), 0);
+    }
+    return confInt.create(confInt.lowerBound(), confInt.upperBound());
```
dkrft: return conInt

the current line of code is just making a copy of the existing confInt, no need to do this

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +680,215 @@ private static void testForBias(
...
+                    .lower(-5)
+                    .upper(-1)
+                    .build();
+    when(noise.computeConfidenceInterval(
+            anyDouble(), anyInt(), anyDouble(), anyDouble(), anyDouble(), anyDouble()))
+            .thenReturn(ConfidenceInterval.create(-5,3));
+    sum.computeResult();
+
+    // The result interval = (-5, 3), but it should be clamped to (-5, 0), because the bounds are negative.
+    assertThat(sum.computeConfidenceInterval(0.152145599))
```
dkrft: using very specific input like 0.152145599 implies some sort of meaning, that in this case does not exist. better to use somnting simple like 0.0, or even better define a constant to reuse and whose name carries menaing such as DEFAULT_NOISED_X = 0.0

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +680,215 @@ private static void testForBias(
...
+
+  @Test
+  public void computeConfidenceInterval_negativeSumBounds() {
+    sum =
+            BoundedSum.builder()
+                    .epsilon(EPSILON)
+                    .delta(DELTA)
+                    .noise(noise)
+                    .maxPartitionsContributed(1)
+                    .lower(-5)
```
dkrft: -5.0 to indicate that this is a double. please also fix in other places

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +680,215 @@ private static void testForBias(
...
+            .thenReturn(ConfidenceInterval.create(-5,3));
+    sum.computeResult();
+
+    // The result interval = (-5, 3) and it should not be clamped.
+    assertThat(sum.computeConfidenceInterval(0.152145599))
+            .isEqualTo(ConfidenceInterval.create(-5, 3));
+  }
+
+  @Test
+  public void computeConfidenceInterval_infiniteSumBounds() {
```
dkrft: Is this really behaviour that we want. I am notsure we need/should be testing for this

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +680,215 @@ private static void testForBias(
...
+                    .delta(0.9)
+                    .noise(noise)
+                    .maxPartitionsContributed(15)
+                    .lower(1)
+                    .upper(5)
+                    .build();
+    sum.addEntry(1);
+    sum.computeResult();
+
+    // The result interval = (-0.5760855815594823, 2.5760855815594823), but it should be clamped to (0, 2.5760855815594823).
```
dkrft: Are these values mathematically exact. The want value should always be exact and computed with a different, external resource, e.g. wolfram alpha. And we should only be testing up to a certain number number of most significant digits since past that, the float value will most likely not be exact.

Considering that the sigma for the gaussian computation is only exact for the 3 most significant digits we should not be testing further than that. I had a similar conversation with Karim yesterday, you can ask him for more information :)

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -678,4 +680,215 @@ private static void testForBias(
...
+                    .noise(noise)
+                    .delta(0.5)
+                    .maxPartitionsContributed(1)
+                    .lower(1)
+                    .upper(10)
+                    .build();
+    sum.addEntry(10);
+    sum.computeResult();
+
+    assertThat(sum.computeConfidenceInterval(0.5)).isEqualTo(ConfidenceInterval.create(3.068528194400547,16.931471805599454));
```
dkrft: Same comment as above

dkrft (changes requested):

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
```diff
@@ -679,4 +687,186 @@ private static void testForBias(
...
+    assertThat(exception)
+            .hasMessageThat()
+            .startsWith("computeResult must be called before calling computeConfidenceInterval.");
+  }
+
+  @Test
+  public void computeConfidenceIntervals_defaultParameters_callsNoiseCorrectly() {
+    sum = getBoundedSumBuilderWithFields().lower(0.0).upper(100.0).build();
+    sum.computeResult();
+    ConfidenceInterval confInt = sum.computeConfidenceInterval(ALPHA);
```
MashaTelyatnikova, line 862: The local variable 'confInt' is never read. You can remove it and leave only "sum.computeConfidenceInterval(ALPHA);"

MashaTelyatnikova (changes requested):

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
laviniamnedelea: Hmm, in the new version i also added _clampsNegativeInterval, so overall what this does is just testing that NEGATIVE_INFINITY gets clamped to 0.0. I am not sure of whether or not i should keep this test.  @MashaTelyatnikova WDYT?

java/tests/com/google/privacy/differentialprivacy/BoundedSumTest.java
laviniamnedelea: Oops, this is like in count, missed that. Thanks!

MashaTelyatnikova (approved):

PR #48, opened 2020-09-02
Golang: Confidence interval for BoundedMean aggregation

kaelhawaty:
- Add confidence interval implementation for BoundedMean

### DIFF
--- a/go/dpagg/mean.go
+++ b/go/dpagg/mean.go
@@ -66,6 +66,9 @@ type BoundedMeanFloat64 struct {
 	resultReturned bool // whether the result has already been returned
 }
 
+// numSplits sets the number of splits used in bruteforce for Confidence intervals.
+var numSplits = 1000
+
 func bmEquallyInitializedFloat64(bm1, bm2 *BoundedMeanFloat64) bool {
 	return bm1.lower == bm2.lower &&
 		bm1.upper == bm2.upper &&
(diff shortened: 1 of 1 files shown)


go/dpagg/mean.go
```diff
@@ -66,6 +66,9 @@ type BoundedMeanFloat64 struct {
 	resultReturned bool // whether the result has already been returned
 }
 
+// numSplits sets the number of splits used in bruteforce for Confidence intervals.
+var numSplits = 1000
```
miracvbasaran, line 70: We could probably inline this: `var numSplits = 1000 // number of splits used in bruteforce for Confidence intervals`

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
 	return clamped
 }
 
+// ComputeConfidenceInterval computes a confidence interval that contains the true mean with a probability
+// greater than or equal to 1 - alpha using the confidence interval for sum and count.
```
miracvbasaran, line 220: We added an extra couple lines here to BoundedSum and Count due to a comment Dennis had, could you please add those here as well?

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+//
+// Note Result() needs to be called before ComputeConfidenceInterval, otherwise this will return an error.
+func (bm *BoundedMeanFloat64) ComputeConfidenceInterval(alpha float64) (noise.ConfidenceInterval, error) {
+	if !bm.resultReturned {
+		return noise.ConfidenceInterval{}, fmt.Errorf("You need to call Result() before calling ComputeConfidenceInterval()")
```
miracvbasaran, line 225: We also changed the error message, again due to a comment Dennis had.

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	}
+	// Brute force for the tightest confidence interval by iterating through
+	// a fixed number of splits numSplits for countAlpha where it takes the
+	// range of values strictly from 0 to alpha.
```
miracvbasaran, line 229: Maybe we could rephrase this:
```
Brute force optimization for finding a tight(er) confidence interval: Iterate
through a fixed numSplits number of splits for countAlpha where it takes the
range of values strictly from 0 to alpha.
```

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	increment := alpha / float64(numSplits)
+	curCountAlpha := increment
+	bestTightness := math.Inf(1)
```
miracvbasaran, line 232: Optional: You could use `tightest` instead of `best` in `bestTightness` and `bestConfInt`.

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+		confInt, err := bm.computeConfidenceIntervalSplit(alpha, curCountAlpha)
+		if err != nil {
+			return noise.ConfidenceInterval{}, err
+		}
+		tightness := confInt.UpperBound - confInt.LowerBound
+		if tightness < bestTightness {
+			bestTightness = tightness
+			bestConfInt = confInt
+		}
+		curCountAlpha += increment
```
miracvbasaran, line 244: You could do this in for, something like
```
for curCountAlpha = increment; curCountAlpha < alpha, curCountAlpha += increment {
```

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	}
+	return bestConfInt, nil
+}
+
+// As opposed to the former function, this function has countAlpha where you can set confidence
+// level for count confidence interval. It computes the coresponding alpha for sum confidence
+// interval to return a confidence interval with a confidence level of at least 1 - meanAlpha.
```
miracvbasaran, line 251: Maybe something like `computeConfidenceIntervalSplit computes a confidence interval for mean with a given 1 - countAlpha confidence level for count. It computes the corresponding alpha for sum that returns a confidence interval with a confidence level of at least 1 - meanAlpha.`

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+//
+// Note Result() needs to be called before ComputeConfidenceInterval, otherwise this will return an error.
```
miracvbasaran, line 253: nit: `computeConfidenceIntervalSplit`

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+func (bm *BoundedMeanFloat64) computeConfidenceIntervalSplit(meanAlpha, countAlpha float64) (noise.ConfidenceInterval, error) {
+	if !bm.resultReturned {
+		return noise.ConfidenceInterval{}, fmt.Errorf("You need to call Result() before calling ComputeConfidenceInterval()")
```
miracvbasaran, line 256: Same comment above the error message.

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	if err != nil {
+		return noise.ConfidenceInterval{}, err
+	}
+	countConfInt, err := bm.count.ComputeConfidenceInterval(countAlpha)
+	if err != nil {
+		return noise.ConfidenceInterval{}, err
+	}
+
+	// The BoundedMean algorithms maximizes bounds in the denominator to 1, since count
+	// cannot be less than one and to avoid division by zero.
```
miracvbasaran, line 273: Well, count can actually be 0 when there are 0 elements. I think we don't need to explain the reasoning here, we can simply say something like `BoundedMean maximizes the denominator (count) to 1.`

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	} else {
+		meanUpperBound = sumConfInt.UpperBound / countConfInt.UpperBound
+	}
+	// We are using normalizedSum so we need to add midPoint to both bounds.
+	meanLowerBound, meanUpperBound = meanLowerBound+bm.midPoint, meanUpperBound+bm.midPoint
+
+	// Clamp mean bounds to lower and upper bounds.
+	meanLowerBound, err = ClampFloat64(meanLowerBound, bm.lower, bm.upper)
+	if err != nil {
+		return noise.ConfidenceInterval{}, err
```
miracvbasaran, line 296: Maybe we could create a more meaningful error message here, something like `fmt.Errorf(Couldn't clamp lower bound for mean: %v, err)`. Otherwise it might be confusing what we are clamping and why.

(Same below)

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	}
+	return noise.ConfidenceInterval{LowerBound: meanLowerBound, UpperBound: meanUpperBound}, nil
+}
+
+func checkSplit(meanAlpha, countAlpha float64) error {
+	if err := checks.CheckAlpha("computeConfidenceIntervalSplit", meanAlpha); err != nil {
+		return err
+	}
+	if 0 >= countAlpha || countAlpha >= meanAlpha {
+		return fmt.Errorf("computeConfidenceIntervalSplit: countAlpha should be strictly between 0 and meanAlpha %f, got %f", meanAlpha, countAlpha)
```
miracvbasaran, line 310: It is preferred to have "got before want", so we could report something like `got countAlpha=%f, should be strictly between 0 and meanAlpha=%f`.

go/dpagg/mean.go
```diff
@@ -213,6 +216,102 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+// Note Result() needs to be called before ComputeConfidenceInterval, otherwise this will return an error.
+func (bm *BoundedMeanFloat64) computeConfidenceIntervalSplit(meanAlpha, countAlpha float64) (noise.ConfidenceInterval, error) {
+	if !bm.resultReturned {
+		return noise.ConfidenceInterval{}, fmt.Errorf("You need to call Result() before calling ComputeConfidenceInterval()")
+	}
+	if err := checkSplit(meanAlpha, countAlpha); err != nil {
+		return noise.ConfidenceInterval{}, err
+	}
+
+	sumAlpha := (meanAlpha - countAlpha) / (1 - countAlpha)
```
miracvbasaran, line 262: The math we are doing here won't work for very small countAlpha. If countAlpha <~ 1e-18, 1 - countAlpha would be 1, giving us a slightly higher sumAlpha => Slightly higher meanAlpha.

I think this should be fine though, so maybe we could comment that it is not exact for tiny countAlpha.

miracvbasaran (changes requested):

go/dpagg/mean.go
kaelhawaty: We had a discussion earlier about this: "If the denominator is much larger than the numerator, then small inaccuracies in the denominator are not noticeable."
For that case, I would assume meanAlpha to be also extremely small, then the above will apply. If on the other hand meanAlpha is close to 1, then shouldn't get better and better as countAlpha gets smaller?

kaelhawaty:
Closing this PR since another PR will be opened including tests.

PR #51, opened 2020-09-11
Golang/Java: Multiple fixes and improvements

kaelhawaty:
Overall fixes and improvements to confidence intervals in Go/Java.
Golang:
    - Add nextSmallerFloat64 and nextLargerFloat64 functions
    - Edit Laplace's confidence interval checks to use checkEpsilonVeryStrict
    - Improvments to confidence interval noise tests

  Java:
    - Mock getSigma computation to return an accurate sigma value for confidence interval gaussian tests.

### DIFF
--- a/go/dpagg/sum_test.go
+++ b/go/dpagg/sum_test.go
@@ -1020,12 +1020,12 @@ func TestSumComputeConfidenceIntervalForInt64Computation(t *testing.T) {
 	}{
 		{
 			desc: "Gaussian",
-			opt:  &BoundedSumInt64Options{Epsilon: 0.1, Delta: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Gaussian())},
+			opt:  &BoundedSumInt64Options{Epsilon: 0.1, Delta: 0.1, Lower: 0, Upper: 5, Noise: noNoise{noise.Gaussian()}},
 			want: noise.ConfidenceInterval{LowerBound: 40, UpperBound: 60},
 		},
 		{
 			desc: "Laplace",
-			opt:  &BoundedSumInt64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Laplace())},
+			opt:  &BoundedSumInt64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: noNoise{noise.Laplace()}},
 			want: noise.ConfidenceInterval{LowerBound: 15, UpperBound: 85},
 		},
 	} {
@@ -1057,13 +1057,13 @@ func TestSumComputeConfidenceIntervalForFloat64Computation(t *testing.T) {
 	}{
 		{
 			desc:      "Gaussian",
-			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Delta: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Gaussian())},
+			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Delta: 0.1, Lower: 0, Upper: 5, Noise: noNoise{noise.Gaussian()}},
 			want:      noise.ConfidenceInterval{LowerBound: 15.898893, UpperBound: 35.101107},
 			tolerance: 1e-3,
 		},
 		{
 			desc:      "Laplace",
-			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Laplace())},
+			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: noNoise{noise.Laplace()}},
 			want:      noise.ConfidenceInterval{LowerBound: 0.0, UpperBound: 60.157359},
 			tolerance: 1e-6,
 		},
@@ -1075,6 +1075,7 @@ func TestSumComputeConfidenceIntervalForFloat64Computation(t *testing.T) {
 		c.Result()
 		got, _ := c.ComputeConfidenceInterval(0.5)
 
+		// TODO: Replace with appropriate approxEqual function.
 		if math.Abs(got.LowerBound-tc.want.LowerBound) > tc.tolerance*math.Max(got.LowerBound, tc.want.LowerBound) {
 			t.Errorf("TestSumComputeConfidenceIntervalForFloat64Computation(Noise: %s)=%0.10f, want %0.10f, LowerBounds are not equal",
 				tc.desc, got.LowerBound, tc.want.LowerBound)
--- a/go/dpagg/sum_test.go
+++ b/go/dpagg/sum_test.go
@@ -1050,19 +1050,22 @@ func TestSumComputeConfidenceIntervalForInt64Computation(t *testing.T) {
 // Tests that ComputeConfidenceInterval for float64 returns a correct interval for a given sum.
 func TestSumComputeConfidenceIntervalForFloat64Computation(t *testing.T) {
 	for _, tc := range []struct {
-		desc string
-		opt  *BoundedSumFloat64Options
-		want noise.ConfidenceInterval
+		desc      string
+		opt       *BoundedSumFloat64Options
+		want      noise.ConfidenceInterval
+		tolerance float64
 	}{
 		{
-			desc: "Gaussian",
-			opt:  &BoundedSumFloat64Options{Epsilon: 0.1, Delta: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Gaussian())},
-			want: noise.ConfidenceInterval{LowerBound: 15.8964252365, UpperBound: 35.1035747635},
+			desc:      "Gaussian",
+			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Delta: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Gaussian())},
+			want:      noise.ConfidenceInterval{LowerBound: 15.898893, UpperBound: 35.101107},
+			tolerance: 1e-3,
 		},
 		{
-			desc: "Laplace",
-			opt:  &BoundedSumFloat64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Laplace())},
-			want: noise.ConfidenceInterval{LowerBound: 0.0000000000, UpperBound: 60.1573590280},
+			desc:      "Laplace",
+			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Laplace())},
+			want:      noise.ConfidenceInterval{LowerBound: 0.0, UpperBound: 60.157359},
+			tolerance: 1e-6,
 		},
 	} {
 		c := NewBoundedSumFloat64(tc.opt)
@@ -1072,11 +1075,11 @@ func TestSumComputeConfidenceIntervalForFloat64Computation(t *testing.T) {
 		c.Result()
 		got, _ := c.ComputeConfidenceInterval(0.5)
 
-		if !ApproxEqual(got.LowerBound, tc.want.LowerBound) {
+		if math.Abs(got.LowerBound-tc.want.LowerBound) > tc.tolerance*math.Max(got.LowerBound, tc.want.LowerBound) {
 			t.Errorf("TestSumComputeConfidenceIntervalForFloat64Computation(Noise: %s)=%0.10f, want %0.10f, LowerBounds are not equal",
 				tc.desc, got.LowerBound, tc.want.LowerBound)
 		}
-		if !ApproxEqual(got.UpperBound, tc.want.UpperBound) {
+		if math.Abs(got.UpperBound-tc.want.UpperBound) > tc.tolerance*math.Max(got.UpperBound, tc.want.UpperBound) {
 			t.Errorf("TestSumComputeConfidenceIntervalForFloat64Computation(Noise: %s)=%0.10f, want %0.10f, UpperBounds are not equal",
 				tc.desc, got.UpperBound, tc.want.UpperBound)
 		}
--- a/go/noise/gaussian_noise.go
+++ b/go/noise/gaussian_noise.go
@@ -123,14 +123,22 @@ func (gaussian) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
 //
 // See https://github.com/google/differential-privacy/tree/main/common_docs/confidence_intervals.md.
 func (gaussian) ComputeConfidenceIntervalInt64(noisedX, l0Sensitivity, lInfSensitivity int64, epsilon, delta, alpha float64) (ConfidenceInterval, error) {
-	err := checkArgsConfidenceIntervalGaussian("computeConfidenceIntervalInt64", l0Sensitivity, float64(lInfSensitivity), epsilon, delta, alpha)
+	err := checkArgsConfidenceIntervalGaussian("ComputeConfidenceIntervalInt64 (gaussian)", l0Sensitivity, float64(lInfSensitivity), epsilon, delta, alpha)
 	if err != nil {
 		err := fmt.Errorf("ComputeConfidenceIntervalInt64(l0sensitivity %d, lInfSensitivity %d, epsilon %f, delta %e, alpha %f) checks failed with %v",
 			l0Sensitivity, lInfSensitivity, epsilon, delta, alpha, err)
 		return ConfidenceInterval{}, err
 	}
 	sigma := SigmaForGaussian(l0Sensitivity, float64(lInfSensitivity), epsilon, delta)
-	return computeConfidenceIntervalGaussian(float64(noisedX), sigma, alpha).roundToInt64(), nil
+	// Computing the confidence interval around zero rather than nosiedX helps represent the
+	// interval bounds more accurately. The reason is that the resolution of float64 values is most
+	// fine grained around zero.
+	confIntAroundZero := computeConfidenceIntervalGaussian(0, sigma, alpha).roundToInt64()
+	// Adding noisedX after converting the interval bounds to int64 ensures that no precision is lost
+	// due to the coarse resolution of float64 values for large instances of noisedX.
+	lowerBound := nextSmallerFloat64(int64(confIntAroundZero.LowerBound) + noisedX)
+	upperBound := nextLargerFloat64(int64(confIntAroundZero.UpperBound) + noisedX)
+	return ConfidenceInterval{LowerBound: lowerBound, UpperBound: upperBound}, nil
 }
 
 // ComputeConfidenceIntervalFloat64 computes a confidence interval that contains the raw value x from which float64
@@ -138,7 +146,7 @@ func (gaussian) ComputeConfidenceIntervalInt64(noisedX, l0Sensitivity, lInfSensi
 //
 // See https://github.com/google/differential-privacy/tree/main/common_docs/confidence_intervals.md.
 func (gaussian) ComputeConfidenceIntervalFloat64(noisedX float64, l0Sensitivity int64, lInfSensitivity, epsilon, delta, alpha float64) (ConfidenceInterval, error) {
-	err := checkArgsConfidenceIntervalGaussian("ComputeConfidenceIntervalFloat64", l0Sensitivity, lInfSensitivity, epsilon, delta, alpha)
+	err := checkArgsConfidenceIntervalGaussian("ComputeConfidenceIntervalFloat64 (gaussian)", l0Sensitivity, lInfSensitivity, epsilon, delta, alpha)
 	if err != nil {
 		err = fmt.Errorf("ComputeConfidenceIntervalFloat64(l0sensitivity %d, lInfSensitivity %f, epsilon %f, delta %e, alpha %f) checks failed with %v",
 			l0Sensitivity, lInfSensitivity, epsilon, delta, alpha, err)
@@ -184,10 +192,15 @@ func addGaussian(x, sigma float64) float64 {
 // computeConfidenceIntervalGaussian computes a confidence interval that contains the raw value x from which
 // float64 noisedX is computed with a probability equal to 1 - alpha with the given sigma.
 func computeConfidenceIntervalGaussian(noisedX, sigma, alpha float64) ConfidenceInterval {
-	z := inverseCDFGaussian(sigma, alpha/2) // z will hold a negative value.
-	FloatLowerBound := noisedX + z
-	FloatUpperBound := noisedX - z
-	return ConfidenceInterval{LowerBound: FloatLowerBound, UpperBound: FloatUpperBound}
+	z := inverseCDFGaussian(sigma, alpha/2)
+	// Because of the symmetry of the Gaussian distribution,
+	// -z corresponds to the (1 - alpha/2)-quantile of the distribution,
+	// meaning that the interval [z, -z] contains 1-alpha of the probability mass.
+	// Deriving the (1 - alpha/2)-quantile from the (alpha/2)-quantile and not vice versa is a
+	// deliberate choice. The reason is that alpha tends to be very small.
+	// Consequently, alpha/2 is more accurately representable as a float64 than 1 - alpha/2,
+	// facilitating numerical computations.
+	return ConfidenceInterval{LowerBound: noisedX + z, UpperBound: noisedX - z}
 }
 
 // inverseCDFGaussian computes the quantile z satisfying Pr[Y <= z] = p for a random variable Y that is Gaussian
--- a/go/noise/gaussian_noise_test.go
+++ b/go/noise/gaussian_noise_test.go
@@ -490,7 +490,7 @@ func TestInverseCDFGaussian(t *testing.T) {
 	} {
 		Zc := inverseCDFGaussian(tc.sigma, tc.p)
 		if !approxEqual(Zc, tc.want) {
-			t.Errorf("TestInverseCDFGaussian(%f, %f) = %0.10f, want %0.10f, desc: %s", tc.sigma, tc.p, Zc, tc.want, tc.desc)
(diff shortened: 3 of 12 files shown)


go/noise/secure_noise_math_test.go
```diff
@@ -257,3 +257,113 @@ func TestRoundToMultipleOfPowerOfTwoXIsNotAMultiple(t *testing.T) {
...
+	want := float64(1 << 63)
+	if got != want {
+		t.Errorf("nextLargerFloat64(%d) = %f, want %f", math.MaxInt64, got, want)
+	}
+}
+
+func TestNextSmallerFloat64InputRepresentableAsFloat64(t *testing.T) {
+	// Verify that nextSmallerFloat64 returns a float64 of the same value.
+	for _, tc := range []struct {
+		n int64
```
MashaTelyatnikova, line 323: I don't think we should use struct here, use ordinary array of int64

go/noise/secure_noise_math_test.go
```diff
@@ -257,3 +257,113 @@ func TestRoundToMultipleOfPowerOfTwoXIsNotAMultiple(t *testing.T) {
 		}
 	}
 }
+
+func TestNextLargerFloat64InputRepresentableAsFloat64(t *testing.T) {
+	// Verify that nextLargerFloat64 returns a float64 of the same value.
+	for _, tc := range []struct {
```
MashaTelyatnikova, line 263: don't use struct, just use int64[]

go/noise/gaussian_noise_test.go
```diff
@@ -612,6 +611,36 @@ func TestComputeConfidenceIntervalInt64Gaussian(t *testing.T) {
...
+			delta:           0.1,
+			alpha:           0.1,
+			// Z value = -5, upperBound = 1 << 58 + 5 = 288230376151711749 gets rounded
+			// down to 288230376151711744, nextLargeFloat64 rounds it back up to 288230376151711808.
+			// lowerBound = 1 << 58 - 5 = 288230376151711739 gets rounded up to 288230376151711744,
+			// nextSmallerFloat64 rounds it to the previous float64 value 288230376151711712.
+			want: ConfidenceInterval{288230376151711712.0, 2.8823037615171181E+17},
+		},
+		{
+			desc:            "High negative noisedX",
```
MashaTelyatnikova, line 630: replace high to large, low to small everywhere in the tests. It is what we decided in the Java tests

MashaTelyatnikova (changes requested):

kaelhawaty:
Rebased to pull new changes and addressed comments.

go/noise/gaussian_noise.go
```diff
@@ -121,20 +121,23 @@ func (gaussian) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
...
 func (gaussian) ComputeConfidenceIntervalInt64(noisedX, l0Sensitivity, lInfSensitivity int64, epsilon, delta, alpha float64) (ConfidenceInterval, error) {
-	err := checkArgsConfidenceIntervalGaussian("computeConfidenceIntervalInt64", l0Sensitivity, float64(lInfSensitivity), epsilon, delta, alpha)
+	err := checkArgsConfidenceIntervalGaussian("ComputeConfidenceIntervalInt64 (gaussian)", l0Sensitivity, float64(lInfSensitivity), epsilon, delta, alpha)
 	if err != nil {
 		err := fmt.Errorf("ComputeConfidenceIntervalInt64(l0sensitivity %d, lInfSensitivity %d, epsilon %f, delta %e, alpha %f) checks failed with %v",
 			l0Sensitivity, lInfSensitivity, epsilon, delta, alpha, err)
 		return ConfidenceInterval{}, err
 	}
 	sigma := SigmaForGaussian(l0Sensitivity, float64(lInfSensitivity), epsilon, delta)
-	return computeConfidenceIntervalGaussian(float64(noisedX), sigma, alpha).roundToInt64(), nil
```
MashaTelyatnikova, line 131: add comment for consistency with Java:
// Computing the confidence interval around zero rather than nosiedX helps represent the
// interval bounds more accurately. The reason is that the resolution of double values is most
// fine grained around zero."

go/noise/gaussian_noise.go
```diff
@@ -121,20 +121,23 @@ func (gaussian) DeltaForThreshold(l0Sensitivity int64, lInfSensitivity, epsilon,
...
+	confIntAroundZero := computeConfidenceIntervalGaussian(0, sigma, alpha).roundToInt64()
+	lowerBound := nextSmallerFloat64(int64(confIntAroundZero.LowerBound) + noisedX)
```
MashaTelyatnikova, line 132: add comment before 132 line:

// Adding noisedX after converting the interval bounds to long ensures that no precision is lost
// due to the coarse resolution of double values for large instances of noisedX.

go/noise/gaussian_noise_test.go
```diff
@@ -1058,14 +1110,14 @@ func TestComputeConfidenceIntervalFloat64Gaussian(t *testing.T) {
 	} {
 		got, err := gauss.ComputeConfidenceIntervalFloat64(tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.delta, tc.alpha)
 		if (err != nil) != tc.wantErr {
-			t.Errorf("ComputeConfidenceIntervalFloat64: when %s got err %v, wantErr=%t", tc.desc, err, tc.wantErr)
+			t.Errorf("ComputeConfidenceIntervalFloat64Gaussian: when %s got err %v, wantErr=%t", tc.desc, err, tc.wantErr)
 		}
 		if !approxEqual(got.LowerBound, tc.want.LowerBound) {
-			t.Errorf("TestComputeConfidenceIntervalFloat64(%f, %d, %f, %f, %f)=%0.10f, want %0.10f, desc %s, LowerBound is not equal",
+			t.Errorf("TestComputeConfidenceIntervalFloat64Gaussian(%f, %d, %f, %f, %f)=%0.10f, want %0.10f, desc %s, LowerBound is not equal",
```
MashaTelyatnikova, line 1116: maybe "lower bounds are not equal"? (the same for upper)

go/noise/laplace_noise.go
```diff
@@ -172,7 +172,10 @@ func (laplace) ComputeConfidenceIntervalInt64(noisedX, l0Sensitivity, lInfSensit
 		return ConfidenceInterval{}, err
 	}
 	lambda := laplaceLambda(l0Sensitivity, float64(lInfSensitivity), epsilon)
-	return computeConfidenceIntervalLaplace(float64(noisedX), lambda, alpha).roundToInt64(), nil
```
MashaTelyatnikova, line 175: // Computing the confidence interval around zero rather than nosiedX helps represent the
// interval bounds more accurately. The reason is that the resolution of double values is most
// fine grained around zero.

go/noise/laplace_noise.go
```diff
@@ -172,7 +172,10 @@ func (laplace) ComputeConfidenceIntervalInt64(noisedX, l0Sensitivity, lInfSensit
...
+	confIntAroundZero := computeConfidenceIntervalLaplace(0, lambda, alpha).roundToInt64()
+	lowerBound := nextSmallerFloat64(int64(confIntAroundZero.LowerBound) + noisedX)
```
MashaTelyatnikova, line 176: before line 176

// Adding noisedX after converting the interval bounds to long ensures that no precision is lost
// due to the coarse resolution of double values for large instances of noisedX.

MashaTelyatnikova (changes requested):

go/noise/gaussian_noise.go
kaelhawaty: Just changed double and long to float64 and int64 respectively in the documentation.

go/noise/gaussian_noise_test.go
kaelhawaty: Changed the error message for other confidence interval tests too in gaussian_noise_test.go

go/noise/gaussian_noise_test.go
```diff
@@ -612,6 +611,36 @@ func TestComputeConfidenceIntervalInt64Gaussian(t *testing.T) {
...
+		},
+		{
+			desc:            "Large negative noisedX",
+			noisedX:         -(1 << 58),
+			l0Sensitivity:   1,
+			lInfSensitivity: 1,
+			epsilon:         0.1,
+			delta:           0.1,
+			alpha:           0.1,
+			// Z value = -5, upperBound = -(1 << 58) + 5 = --288230376151711739 gets rounded
```
dkrft, line 637: nit: -288230376151711739

go/noise/gaussian_noise_test.go
```diff
@@ -612,6 +611,36 @@ func TestComputeConfidenceIntervalInt64Gaussian(t *testing.T) {
...
+			// down to -288230376151711744, nextLargeFloat64 rounds it back up to 288230376151711712.
```
dkrft, line 638: nit: -288230376151711712

go/noise/gaussian_noise_test.go
```diff
@@ -856,6 +896,30 @@ func TestComputeConfidenceIntervalFloat64Gaussian(t *testing.T) {
 			want:            ConfidenceInterval{-96.6140883158, 236.6140883158},
 			wantErr:         false,
 		},
+		// float64 precision should be accurate for abs(noisedX) < 2^54
```
dkrft, line 899: I am not sure I understand the purpose of this comment.

go/noise/gaussian_noise_test.go
```diff
@@ -856,6 +896,30 @@ func TestComputeConfidenceIntervalFloat64Gaussian(t *testing.T) {
...
+		{
+			desc:            "Large positive noisedX",
+			noisedX:         38475693,
```
dkrft, line 902: IIUC, this is actually a float value. It makes your code easier to read and understand if you signify integral float values by writing them as 38475693.0.

Please also fix in other places.

go/noise/gaussian_noise_test.go
```diff
@@ -445,19 +445,19 @@ func TestInverseCDFGaussian(t *testing.T) {
 	}{
 
 		{
-			desc:  "Abitrary input test",
+			desc:  "Arbitrary input test",
 			sigma: 1,
 			p:     0.05,
 			want:  -1.64485362695,
```
dkrft, line 451: Approx equal is only considers the 6 most significant digits. Having a want value of the form -1.64485362695 implies much greater accuracy than what is actually offered by the test. It may make sense to note the error tolerance of the test and prune the want values accordingly (unless you think the error tolerance will be adjusted later on)

Also, I was wondering how the want values are computed. They should not be computed by the code we are testing but via an independent method, e.g., Wolfram Alpha. When copmuting them make sure that they are truely exact up to the digits you are providing (Wolfram Alpha should be able to give you enough precision for that).

If necessary, please also fix for other Gaussian and Laplace tests.

On a side note, the sigma computed by the Gaussian noise is only accurate up to the 3 most significant digits. Thus, whenever your test something that implicitly uses a sigma computed by Gaussian noise, you can only test thefor the 3 most significant digits. To get an accurate want value, you should first compute a more accurate sigma (e.g. by ramping up the respective accuracy parameter of Gaussian noise) and then use Wolfram Alpha or something similar to get the respect bounds associated with this sigma.

dkrft (changes requested):

go/noise/gaussian_noise_test.go
kaelhawaty: This was mainly to test that high noised values can be presented accurately in the float64 version of the function as long as the noisedX < 2^54

go/noise/gaussian_noise_test.go
kaelhawaty: Yeah that is something I also wanted to change, perhaps a approxEqual function with Tolerance as a parameter is better.

For Laplace, The math is simple so I computed every step of the process using a high precision calculator (up to 22 digits). In laplace_noise_test_go, some of the cases has extra significant digits because I think the error tolerance should be adjusted accordingly for those values. I think the same case applies in gaussian_noise_test.go

For Gaussian for the 4 tests I added, I have used SigmaForGaussian function to compute the corresponding sigma and then calculate everything from there just as Laplace. I think this makes sense in terms we are testing the accuracy of the calculations for confidence interval not the accuracy of the sigma parameter. @laviniamnedelea can also provide her way of coming up with want values for Gaussian.

I can maybe change the gaussian test by computing a more accurate sigma and then recalculating all the want values, however I think the confidence interval will change significantly in some tests. Moreover, I think these tests should be only concerned with confidence interval calculations like I said previously.

The only problem I see with decrease the tolerance to something lower than 1e-6 (for ex 1e-10) is the inaccuracy that might result in Gaussian confidence interval computations since we use erfcInv function so I need to check how many digits are guaranteed to be accurate for gaussian.

go/noise/gaussian_noise_test.go
dkrft: I see. Makes sense. Maybe something like "testing that bounds are accurate for abs(bound) < 2^53" is more descriptive.

go/noise/gaussian_noise_test.go
dkrft: I wrote some additional tests for Java and it was accurate for the 3 most significant digits (but not more accurate, but the reason was most likely the sigma computation)

Happy to hear that you did independent validation. For the sigma cases, I think the right thing to do is compute a sigma that has the appropriate accuracy (unfrotunately, the only way I know how to do this easily is using the library with a bit more accuracy) and then use that accurate sigma in the manual coputation. Testing with 1e-3 tolerance should still accept. If not, we know there is a problem. (I can also send you some code from the Java library for reference (The context here is that we made computeQuantile part of the interface, so I added some extra tests for this)

kaelhawaty:
This was mainly to fix the issue for accuracy of sigma parameter calculation. This is to make the confidence interval tests independent of the accuracy of the getSigma function
https://github.com/google/differential-privacy/blob/bc8eeaccb233947691f7ce104f00658f72deff63/java/main/com/google/privacy/differentialprivacy/GaussianNoise.java#L232

getSigma ensures at most 3 significant digits for the sigma parameter however this doesn't map to a 3 significant digits accuracy for confidence intervals.

Consider one example:
noisedX = 100, Z-value (quantile) = 100.xxx (where xxx means inaccurate digits)
In computing for lowerBound:
noisedX - Z-value = 100 - 100.xxx = 0.xxx 
Thus we end up with a result that is not accurate to any digit. For most of the calculations, we can't guarantee 3 significant digits.

For Golang:
- Go doesn't support mocking in a simple way so the way it is fixed there is since we are already have tests for 

https://github.com/google/differential-privacy/blob/bc8eeaccb233947691f7ce104f00658f72deff63/go/noise/gaussian_noise.go#L182

This function calculates confidence interval with sigma as a parameter so testing this function is also sufficient for the float64 version of the interface function ComputeConfidenceIntervalFloat64()

https://github.com/google/differential-privacy/blob/bc8eeaccb233947691f7ce104f00658f72deff63/go/noise/gaussian_noise.go#L123

For ComputeConfidenceIntervalInt64(), 3 Significant digits is enough for the tests that is already in the library. The computation with a more exact sigma parameter gives out the same result because of rounding.

Java:
- I have changed the visibility of getSigma to package private and mocked the function to return an exact sigma parameter to 8 significant digits. This way Confidence interval tests are independent of the accuracy of the function.

All the want values also has been rounded to match the current tolerance for better readability and to ensure all the digits are tested.

kaelhawaty:
Rebased on the main branch to pull new changes.

go/noise/gaussian_noise_test.go
```diff
@@ -795,101 +818,47 @@ func TestComputeConfidenceIntervalInt64Gaussian(t *testing.T) {
...
+		if (got != ConfidenceInterval{}) {
+			t.Errorf("TestComputeConfidenceIntervalInt64GaussianArgumentCheck(%d, %d, %d, %f, %f)=[LowerBound: %0.10f, UpperBound: %0.10f], want [LowerBound: 0.0, UpperBound: 0.0]",
+				tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.alpha, got.LowerBound, got.UpperBound)
 		}
 	}
 }
 
-func TestComputeConfidenceIntervalFloat64Gaussian(t *testing.T) {
+func TestComputeConfidenceIntervalFloat64GaussianArgumentCheck(t *testing.T) {
+	// Test
```
MashaTelyatnikova, line 836: do we need this comment? doesn't add much info :)

go/noise/gaussian_noise_test.go
```diff
@@ -1052,21 +993,16 @@ func TestComputeConfidenceIntervalFloat64Gaussian(t *testing.T) {
...
-		if !approxEqual(got.LowerBound, tc.want.LowerBound) {
-			t.Errorf("TestComputeConfidenceIntervalFloat64(%f, %d, %f, %f, %f)=%0.10f, want %0.10f, desc %s, LowerBound is not equal",
-				tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.alpha, got.UpperBound, tc.want.UpperBound, tc.desc)
+		if err == nil {
+			t.Errorf("ComputeConfidenceIntervalFloat64Gaussian: when %s no error was returned, expected error", tc.desc)
 		}
-		if !approxEqual(got.UpperBound, tc.want.UpperBound) {
-			t.Errorf("TestComputeConfidenceIntervalFloat64(%f, %d, %f, %f, %f)=%0.10f, want %0.10f, desc %s, UpperBound is not equal",
-				tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.alpha, got.LowerBound, tc.want.LowerBound, tc.desc)
+		// Checks default confidence interval is returned when err is generated.
```
MashaTelyatnikova, line 1002: we don't check this value. It can be easily changed. THe only thing we need to know is that there were an error.
Please fix in other places, if you have similar one.

go/noise/laplace_noise_test.go
```diff
@@ -464,22 +437,18 @@ func TestComputeConfidenceIntervalFloat64Laplace(t *testing.T) {
...
-				tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.alpha, got.LowerBound, tc.want.LowerBound, tc.desc)
+
+		if err == nil {
+			t.Errorf("ComputeConfidenceIntervalFloat64Laplace: when %s no error was returned, expected error", tc.desc)
 		}
-		if !approxEqual(got.UpperBound, tc.want.UpperBound) {
-			t.Errorf("TestComputeConfidenceIntervalFloat64Laplace(%f, %d, %f, %f, %f)=%0.10f, want %0.10f, desc %s, UpperBounds are not equal",
-				tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.alpha, got.UpperBound, tc.want.UpperBound, tc.desc)
+		// Checks default confidence interval is returned when err is generated.
+		if (got != ConfidenceInterval{}) {
```
MashaTelyatnikova, line 449: the same

go/noise/laplace_noise_test.go
```diff
@@ -640,22 +625,18 @@ func TestComputeConfidenceIntervalInt64Laplace(t *testing.T) {
...
-				tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.alpha, got.LowerBound, tc.want.LowerBound, tc.desc)
+
+		if err == nil {
+			t.Errorf("ComputeConfidenceIntervalInt64Laplace: when %s no error was returned, expected error", tc.desc)
 		}
-		if got.UpperBound != tc.want.UpperBound {
-			t.Errorf("TestComputeConfidenceIntervalInt64Laplace(%d, %d, %d, %f, %f)=%0.10f, want %0.10f, desc %s, UpperBounds are not equal",
-				tc.noisedX, tc.l0Sensitivity, tc.lInfSensitivity, tc.epsilon, tc.alpha, got.UpperBound, tc.want.UpperBound, tc.desc)
+		// Checks default confidence interval is returned when err is generated.
+		if (got != ConfidenceInterval{}) {
```
MashaTelyatnikova, line 637: the same

go/noise/secure_noise_math.go
```diff
@@ -64,3 +64,38 @@ func ceilPowerOfTwo(x float64) float64 {
...
+// down the int64 value to the nearest representable float64. This function ensures that int64 n <=
+// float64 nextLargerFloat64(n)
+func nextLargerFloat64(n int64) float64 {
+	// Large int64 values n may lie between two representable float64 values a and b,
+	// i.e., a < n < b, (note that in this case a and b are guaranteed to be integers).
+	// If the standard conversion to float64 rounds the int64 value down, e.g. float64(n) = a,
+	// the difference a - n will be negative, indicating that the result needs to be incremented
+	// to the next float64 value b.
+	result := float64(n)
+	dif := int64(result) - n
```
MashaTelyatnikova, line 81: better "diff"

go/noise/secure_noise_math.go
```diff
@@ -64,3 +64,38 @@ func ceilPowerOfTwo(x float64) float64 {
...
+	if dif < 0 {
+		return math.Nextafter(result, math.Inf(1))
```
MashaTelyatnikova, line 83: shouldn't this name be "NextAfter"?

go/dpagg/sum_test.go
```diff
@@ -1050,19 +1050,22 @@ func TestSumComputeConfidenceIntervalForInt64Computation(t *testing.T) {
...
+			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Delta: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Gaussian())},
+			want:      noise.ConfidenceInterval{LowerBound: 15.898893, UpperBound: 35.101107},
+			tolerance: 1e-3,
 		},
 		{
-			desc: "Laplace",
-			opt:  &BoundedSumFloat64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Laplace())},
-			want: noise.ConfidenceInterval{LowerBound: 0.0000000000, UpperBound: 60.1573590280},
+			desc:      "Laplace",
+			opt:       &BoundedSumFloat64Options{Epsilon: 0.1, Lower: 0, Upper: 5, Noise: getNoiselessConfInt(noise.Laplace())},
```
MashaTelyatnikova, line 1066: Noise: noNoise{noise.Laplace()} maybe?

getNoislessConfInt is not a great name, you return not a confidence interval, but noise (so either not use extra method or rename it)

go/dpagg/sum_test.go
```diff
@@ -1072,11 +1075,11 @@ func TestSumComputeConfidenceIntervalForFloat64Computation(t *testing.T) {
 		c.Result()
 		got, _ := c.ComputeConfidenceInterval(0.5)
 
-		if !ApproxEqual(got.LowerBound, tc.want.LowerBound) {
+		if math.Abs(got.LowerBound-tc.want.LowerBound) > tc.tolerance*math.Max(got.LowerBound, tc.want.LowerBound) {
```
MashaTelyatnikova, line 1078: why we can't use ApproxEqual?

go/noise/gaussian_noise_test.go
```diff
@@ -1052,21 +993,16 @@ func TestComputeConfidenceIntervalFloat64Gaussian(t *testing.T) {
...
+		if (got != ConfidenceInterval{}) {
+			t.Errorf("TestComputeConfidenceIntervalFloat64GaussianArgumentCheck(%f, %d, %f, %f, %f)=[LowerBound: %0.10f, UpperBound: %0.10f], want [LowerBound: 0.0, UpperBound: 0.0]",
```
MashaTelyatnikova, line 1004: TestComputeConfidenceIntervalFloat64GaussianArgumentCheck - instead of the test name (which will anyway appear in the log) you should put tthe name of the function you are testing, in this scenario it should be "ComputeConfidenceIntervalFloat64(****)"


check this rule for other tests please

MashaTelyatnikova (changes requested):

go/noise/secure_noise_math.go
kaelhawaty: It should but it is a function in the Go math package: https://godoc.org/math#Nextafter

go/dpagg/sum_test.go
kaelhawaty: Yeah makes sense, This function was added a month ago though. The reason to have a function in the first place is if you don't initialize noNoise struct with an actual noise object, you will get a Go's version of "nullpointerexception" so I made this function as a sort of constructor to force that you need to initialize it with either laplace or gaussian.
https://github.com/google/differential-privacy/blob/2ccce87b6f4508efd991732f7227dcda3cd39fb2/go/dpagg/dpagg_test.go#L58-L70

I am not sure of another name that conveys this "get a Noiseless interface for confidence intervals using X distribution".

go/dpagg/sum_test.go
kaelhawaty: Gaussian has only at most 3 significant digits accuracy thus it needs a tolerance of 1e-3 while Laplace is as accurate as it gets. Now because approxEqual doesn't have a parameter for tolerance and this is the only test which needs one, I have written it in the if condition.

MashaTelyatnikova: I see, I would still add new method which accepts tolerance as a parameter (might be useful in the future) and the old method "ApproxEqual" I would rewrite - so it would call our new method with default tolerance. WDYT?

go/noise/secure_noise_math.go
MashaTelyatnikova: Oh, sorry, didn't notice. Yes, we can resolve this my comment.

MashaTelyatnikova (changes requested):

go/dpagg/sum_test.go
kaelhawaty: Removed the function and replaced it with noNoise{noise} in each reference.

go/dpagg/sum_test.go
kaelhawaty: Added TODO so that these test utilities can be fixed for all packages later.

kaelhawaty:
Closing this PR to split into 2 smaller PR so that it is more easily reviewed.

PR #52, opened 2020-09-14
Golang: Confidence interval for BoundedMean aggregation.

laviniamnedelea:


go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	// range of values strictly from 0 to alpha.
+	increment := alpha / float64(numSplits)
+	tightest := math.Inf(1)
+	var tightestConfInt noise.ConfidenceInterval
+	for curCountAlpha := increment; curCountAlpha < alpha; curCountAlpha += increment {
+		confInt, err := bm.computeConfidenceIntervalSplit(alpha, curCountAlpha)
+		if err != nil {
+			return noise.ConfidenceInterval{}, err
+		}
+		tightness := confInt.UpperBound - confInt.LowerBound
```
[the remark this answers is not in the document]
MashaTelyatnikova: Also, I am not sure, but will it make sense to use absolute value here? math.Abs(confInt.UpperBound - confInt.LowerBound)?

go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
 	return clamped
 }
 
+// ComputeConfidenceInterval computes a confidence interval that contains the true mean with
+// probability greater than or equal to 1 - alpha. The computation is based exclusively on the
```
MashaTelyatnikova, line 219: should we say maybe that count and sum based only on the noised value and privacy params?
Because from this sentence (and if reader doesn't know details) it is not clear why this leads to not using extra budget.

go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+}
+
+// computeConfidenceIntervalSplit computes a confidence interval for mean with a given 1 - countAlpha
+// confidence level for count. It computes the corresponding alpha for sum to return a confidence interval
+// with a confidence level of at least 1 - meanAlpha. The computation is based exclusively on the confidence
+// intervals for sum and count. Thus no privacy budget is consumed by this operation.
+//
+// Result() needs to be called before ComputeConfidenceInterval, otherwise this will return an error.
+func (bm *BoundedMeanFloat64) computeConfidenceIntervalSplit(meanAlpha, countAlpha float64) (noise.ConfidenceInterval, error) {
+	if !bm.resultReturned {
```
MashaTelyatnikova, line 254: I am not sure why we need this check here? This is a private method and the "main" method which calls it already has this check.

go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+		meanLowerBound = sumConfInt.LowerBound / countConfInt.UpperBound
+	} else {
+		meanLowerBound = sumConfInt.LowerBound / countConfInt.LowerBound
+	}
+	if sumConfInt.UpperBound >= 0 {
+		meanUpperBound = sumConfInt.UpperBound / countConfInt.LowerBound
+	} else {
+		meanUpperBound = sumConfInt.UpperBound / countConfInt.UpperBound
+	}
+	// We are using normalizedSum so we need to add midPoint to both bounds.
```
MashaTelyatnikova, line 288: can you maybe refer to the original formulas or at least algorithm at the Result() method?

go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	meanLowerBound, meanUpperBound = meanLowerBound+bm.midPoint, meanUpperBound+bm.midPoint
+
+	// Clamp mean bounds to lower and upper bounds.
```
MashaTelyatnikova, line 291: can you add comment why?

go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+		return noise.ConfidenceInterval{}, fmt.Errorf("Couldn't clamp upper bound for mean: %v", err)
+	}
+	return noise.ConfidenceInterval{LowerBound: meanLowerBound, UpperBound: meanUpperBound}, nil
+}
+
+func checkSplit(meanAlpha, countAlpha float64) error {
+	if err := checks.CheckAlpha("computeConfidenceIntervalSplit", meanAlpha); err != nil {
+		return err
+	}
+	if 0 >= countAlpha || countAlpha >= meanAlpha {
```
MashaTelyatnikova, line 307: very nit: can you make it "countAlpha <= 0"?

go/dpagg/mean_test.go
```diff
@@ -738,3 +738,153 @@ func TestBMFloat64Serialization(t *testing.T) {
...
+				tc.meanOpt.Lower, tc.meanOpt.Upper, got.UpperBound, tc.want.UpperBound)
+		}
+	}
+}
+
+/**
+ * This test was designed to be not deterministic. It goes along with deterministic analogues in
+ * order to ensure that they don't miss something.
+ */
+func TestMeanComputeConfidenceInterval_confidenceBoundsAlwaysInsideProvidedBoundaries(t *testing.T) {
```
dkrft, line 803: It looks like this test is randomizing anything it can get its hands on. Bt this makes it hard to understand what is happening If it throws an error and to reproduce. The test claims it wants to check that bounding is done correctly. To thest this, one deterministic test should be enough simly define a some lower bound of e.g. 0.0, upper bound of e.g. 1.0 and add interesting values -1.0, 0.0, 1.0 and 2.0

go/dpagg/mean_test.go
```diff
@@ -738,3 +738,153 @@ func TestBMFloat64Serialization(t *testing.T) {
...
+			if meanConfInt.UpperBound < lower || meanConfInt.UpperBound > upper {
+				t.Errorf("TestMeanComputeConfidenceInterval_confidenceBoundsAlwaysInsideProvidedBoundaries")
+			}
+
+		}
+	}
+}
+
+/**
+ * This test was designed to be not deterministic. Brute force for optimal ratio should always
```
dkrft, line 831: A general guideline. Don't do random unit tests just becuase you can't think of plausible test parameters or because you think you will get more test coverage.

The only reason to have randomness in your unit test is if you want to test a statistical property. And in that case you should have clear bounds on the success/failure probability of the test

dkrft (changes requested):

go/dpagg/mean.go
laviniamnedelea: "The computation is based exclusively on the confidence intervals for sum and count, which are based only on the noised value and privacy params" is this ok?

go/dpagg/mean.go
laviniamnedelea: Yep, you are right, this check is not needed.

go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	if !bm.resultReturned {
+		return noise.ConfidenceInterval{}, fmt.Errorf("Result() must be called before calling ComputeConfidenceInterval()")
+	}
+	// Brute force optimization for finding the tightest confidence interval: Iterate
+	// through a fixed numSplits number of splits for countAlpha where it takes the
+	// range of values strictly from 0 to alpha.
+	increment := alpha / float64(numSplits)
+	tightest := math.Inf(1)
+	var tightestConfInt noise.ConfidenceInterval
+	for curCountAlpha := increment; curCountAlpha < alpha; curCountAlpha += increment {
```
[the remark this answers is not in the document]
laviniamnedelea: Sounds good to me!

go/dpagg/mean.go
```diff
@@ -213,6 +215,101 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+		return noise.ConfidenceInterval{}, fmt.Errorf("Couldn't clamp lower bound for mean: %v", err)
+	}
+	meanUpperBound, err = ClampFloat64(meanUpperBound, bm.lower, bm.upper)
+	if err != nil {
+		return noise.ConfidenceInterval{}, fmt.Errorf("Couldn't clamp upper bound for mean: %v", err)
+	}
+	return noise.ConfidenceInterval{LowerBound: meanLowerBound, UpperBound: meanUpperBound}, nil
+}
+
+func checkSplit(meanAlpha, countAlpha float64) error {
```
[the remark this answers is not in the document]
laviniamnedelea: Hmm, i thought adding it would be a good assurance, but i think it's right that it might be unnecessary. Should i remove it altogether, @MashaTelyatnikova WDYT?

go/dpagg/mean_test.go
laviniamnedelea: I understand, therefore would it be best to set some specific bounds and add specific numbers like the previous test? (only one deterministic test)

go/dpagg/mean.go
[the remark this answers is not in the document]
MashaTelyatnikova: +1 for removing it

go/dpagg/mean.go
MashaTelyatnikova: Now I think this comment is redundant tbh, I would remove it at all

go/dpagg/mean.go
MashaTelyatnikova: parameters.

MashaTelyatnikova (changes requested):

go/dpagg/mean_test.go
dkrft: Yes, specific bounds and parametrs would be better 👍

go/dpagg/mean_test.go
```diff
@@ -738,3 +738,109 @@ func TestBMFloat64Serialization(t *testing.T) {
...
+// ComputeConfidenceIntervalInt64 checks that the parameters passed are the ones we expect for count.
+func (mn mockBMNoise) ComputeConfidenceIntervalInt64(noisedX, l0, lInf int64, eps, del, alpha float64) (noise.ConfidenceInterval, error) {
+	return noise.ConfidenceInterval{}, nil
+}
+
+func TestMeanComputeConfidenceInterval_callsNoiseComputeConfidenceIntervalCorrectly(t *testing.T) {
+	bmf := getMockBMF(t)
+	bmf.Result()
+	bmf.computeConfidenceIntervalForExplicitAlphaSplit(alphaLevel, alphaLevel/2)
+}
```
kaelhawaty, line 838: I am not sure if this tests callsNoiseComputeConfidenceIntervalCorrectly since functions on Lines 825 and 830 are empty.

Refer to how this was done in the same file for addNoise: https://github.com/google/differential-privacy/blob/b8aacd8369c4a3a7d3ed0872231530edfdeecde7/go/dpagg/mean_test.go#L237

Also here is an example of how it was done in Count: [Example](https://github.com/google/differential-privacy/blob/514d8c56e7c1226116e08168a8e5bd8270767663/go/dpagg/count_test.go#L478-L531)

go/dpagg/mean_test.go
```diff
@@ -738,3 +738,143 @@ func TestBMFloat64Serialization(t *testing.T) {
...
+func TestMeanComputeConfidenceIntervalForExplicitAlphaSplit(t *testing.T) {
+	noNoise := noNoise{} // To skip initial argument checking.
+	for _, tc := range []struct {
+		meanOpt      *BoundedMeanFloat64Options
+		sumConfInt   noise.ConfidenceInterval
+		countConfInt noise.ConfidenceInterval
+		want         noise.ConfidenceInterval // Confidence Interval after Post-processing.
+	}{
+		{ // Positive lower and upper bounds for sum.
+			meanOpt:      &BoundedMeanFloat64Options{Lower: 0.2, Upper: 15.0, Noise: noNoise, MaxContributionsPerPartition: 1},
```
MashaTelyatnikova, line 751: re-format this options like below

	{"default options", &BoundedMeanFloat64Options{
			Epsilon:                      ln3,
			Lower:                        0,
			Upper:                        1,
			Delta:                        0,
			MaxContributionsPerPartition: 1,
		}},

MashaTelyatnikova: in other tests below as well

MashaTelyatnikova (changes requested):

go/dpagg/mean.go
```diff
@@ -213,6 +213,80 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+	confIntNum, err := bm.normalizedSum.ComputeConfidenceInterval(alphaDen)
+	if err != nil {
+		return noise.ConfidenceInterval{}, err
+	}
+	confIntDen, err := bm.count.ComputeConfidenceInterval(alphaNum)
+	if err != nil {
+		return noise.ConfidenceInterval{}, err
+	}
+
+// Ensuring that the lower and upper bounds of the denominator are consistent with how Result() processes the denominator.
```
MashaTelyatnikova, line 261: can you add indentation to this comment? to the column there next line starts?

go/dpagg/mean.go
```diff
@@ -213,6 +213,80 @@ func (bm *BoundedMeanFloat64) Result() float64 {
...
+		meanLowerBound = confIntNum.LowerBound / confIntDen.UpperBound
+	} else {
+		meanLowerBound = confIntNum.LowerBound / confIntDen.LowerBound
+	}
+	if confIntNum.UpperBound >= 0 {
+		meanUpperBound = confIntNum.UpperBound / confIntDen.LowerBound
+	} else {
+		meanUpperBound = confIntNum.UpperBound / confIntDen.UpperBound
+	}
+// Ensuring that the lower and upper bounds of the mean are consistent with how Reult() processes the mean.
```
MashaTelyatnikova, line 276: the same

MashaTelyatnikova (changes requested):

MashaTelyatnikova (approved):

PR #54, opened 2020-09-29
Java: Confidence Interval for BoundedMean

kaelhawaty:


java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -70,6 +71,20 @@ private static void mockLongNoise(Noise noise, long value) {
     when(noise.getMechanismType()).thenReturn(MechanismType.GAUSSIAN);
   }
 
+  private static void mockDoubleConfInt(Noise noise, ConfidenceInterval confInt) {
```
MashaTelyatnikova, line 74: I am not sure we need to pass noise. Someone can pass actual noise implementation and it will break everything.

Also why we set LAPLACE here? when in setup we use gauissian?

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -372,6 +387,157 @@ public void computeResult_resultAlwaysInsideProvidedBoundaries() {
...
+    mean.addEntry(7.0);
+    mean.computeResult();
+    mean.computeConfidenceInterval(alpha, alpha / 2);
+
+    // Confidence interval for normalized sum.
+    verify(noise)
+        .computeConfidenceInterval(
+            eq(/* x1 + x2 - midpoint * count = 3 + 7 - 6 * 2 = */ -2.0),
+            eq(maxPartitionsContributed),
+            eq(/* maxContributionsPerPartition * (upper - lower) / 2 = 3 * (10 - 2) / 2 = */ 12.0),
```
MashaTelyatnikova, line 422: lInfSensitivity = maxContributionsPerPartition * (upper - lower) / 2 = 3 * (10 - 2) / 2 =

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -372,6 +387,157 @@ public void computeResult_resultAlwaysInsideProvidedBoundaries() {
...
+    mean.addEntry(3.0);
+    mean.addEntry(7.0);
+    mean.computeResult();
+    mean.computeConfidenceInterval(alpha, alpha / 2);
+
+    // Confidence interval for normalized sum.
+    verify(noise)
+        .computeConfidenceInterval(
+            eq(/* x1 + x2 - midpoint * count = 3 + 7 - 6 * 2 = */ -2.0),
+            eq(maxPartitionsContributed),
```
MashaTelyatnikova, line 421: /* l0Sensitivity = */

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -372,6 +387,157 @@ public void computeResult_resultAlwaysInsideProvidedBoundaries() {
...
+
+    mean.addEntry(3.0);
+    mean.addEntry(7.0);
+    mean.computeResult();
+    mean.computeConfidenceInterval(alpha, alpha / 2);
+
+    // Confidence interval for normalized sum.
+    verify(noise)
+        .computeConfidenceInterval(
+            eq(/* x1 + x2 - midpoint * count = 3 + 7 - 6 * 2 = */ -2.0),
```
MashaTelyatnikova, line 420: /* x = x1 + x2 - midpoint * count = 3 + 7 - 6 * 2 = */

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -372,6 +387,157 @@ public void computeResult_resultAlwaysInsideProvidedBoundaries() {
...
+            eq(EPSILON / 2.0),
+            eq(DELTA / 2.0),
+            eq(alpha / 2));
+
+    // Confidence interval for count.
+    verify(noise)
+        .computeConfidenceInterval(
+            eq(/* count */ 2L),
+            eq(maxPartitionsContributed),
+            eq(/* sensitivity of count  = maxContributionsPerPartition */ (long) maxContributionsPerPartition),
```
MashaTelyatnikova, line 432: same comments as above, and non need to specify that this is for count, line 427 says that

MashaTelyatnikova (changes requested):

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
kaelhawaty: I have set it to GAUSSIAN instead.
To remove the noise parameter for that reason, maybe edit this as well ?
https://github.com/google/differential-privacy/pull/54/files#diff-4b1a707aed1d5c6296064d1daa2ee6beR62-R71

MashaTelyatnikova: do we need to set it to GAUSSIAN here if it was set in setUp method?
I would remove it , as well as the noise parameter (in both mock**Noise and mock**ConfInt)

MashaTelyatnikova (changes requested):

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
kaelhawaty: We only need to set it up once in the setUp method since it is required in checks
https://github.com/google/differential-privacy/blob/2ccce87b6f4508efd991732f7227dcda3cd39fb2/java/main/com/google/privacy/differentialprivacy/BoundedMean.java#L246
and implicitly in BoundedSum:
https://github.com/google/differential-privacy/blob/2ccce87b6f4508efd991732f7227dcda3cd39fb2/java/main/com/google/privacy/differentialprivacy/BoundedSum.java#L384

I removed the noise parameter, made the mock object static and just set it once in setUp method.

java/main/com/google/privacy/differentialprivacy/BoundedMean.java
```diff
@@ -175,6 +175,74 @@ public double computeResult() {
...
+  }
+
+  /**
+   * See {@link #computeConfidenceInterval(double)}.
+   *
+   * <p>As opposed to the former function, this function computes a confidence interval for mean
+   * with the additional constraint that the confidence level of the mean's numerator is {@code 1 -
+   * alphaNum}.
+   */
+  ConfidenceInterval computeConfidenceInterval(double alpha, double alphaNum) {
```
MashaTelyatnikova, line 215: why package private?
Can't it be private? If it can then we should fix this, since the good practice is to use as low access level as possible

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -54,27 +54,41 @@
...
 
   @Before
   public void setUp() {
+    when(noise.getMechanismType()).thenReturn(MechanismType.GAUSSIAN);
     // Mock the noise mechanism so that it does not add any noise.
-    mockDoubleNoise(noise, 0);
-    mockLongNoise(noise, 0);
+    mockDoubleNoise(0);
+    mockLongNoise(0);
+    mockDoubleConfInt(ConfidenceInterval.create(0,0));
```
MashaTelyatnikova, line 90: add space in "0,0" statements

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -217,9 +231,9 @@ public void computeResult_callsNoiseCorrectly() {
   @Test
   public void computeResult_addsNoiseToSum() {
     // Mock the noise mechanism so that it adds noise to the sum == 10.0.
-    mockDoubleNoise(noise, 10);
+    mockDoubleNoise(10);
     // Mock the noise mechanism so that it adds noise to the count == 0.
-    mockLongNoise(noise, 0);
+    mockLongNoise(0);
```
MashaTelyatnikova, line 236: is it necessary? It should be 0 since it was in a setup method. IIUC setup is called every time before executing each test

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -244,9 +258,9 @@ public void computeResult_addsNoiseToSum() {
   @Test
   public void computeResult_addsNoiseToCount() {
     // Mock the noise mechanism so that it adds noise to the sum == 0.0.
-    mockDoubleNoise(noise, 0);
+    mockDoubleNoise(0);
```
MashaTelyatnikova, line 261: the same

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -274,9 +288,9 @@ public void computeResult_clampsTooHighAverage() {
     // (i.e., outside of the bounds). If no noise is added then the average will always be within
     // the bounds because the input values are clamped.
     // The noise added to sum is 100.
-    mockDoubleNoise(noise, 100);
+    mockDoubleNoise(100);
     // The noise added to sum is 0.
-    mockLongNoise(noise, 0);
+    mockLongNoise(0);
```
MashaTelyatnikova, line 293: the same

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -305,9 +319,9 @@ public void computeResult_clampsTooLowAverage() {
     // (i.e., outside of the bounds). If no noise is added then the average will always be within
     // the bounds because the input values are clamped.
     // The noise added to sum is 100.
-    mockDoubleNoise(noise, -100);
+    mockDoubleNoise(-100);
     // The noise added to sum is 0.
-    mockLongNoise(noise, 0);
+    mockLongNoise(0);
```
MashaTelyatnikova, line 324: the same

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -372,6 +386,157 @@ public void computeResult_resultAlwaysInsideProvidedBoundaries() {
...
   }
 
+  /** Calling computeConfidenceInterval before noised mean is computed throws an exception. */
+  @Test
+  public void computeConfidenceInterval_calledBeforeResult() {
+    assertThrows(IllegalStateException.class, () -> mean.computeConfidenceInterval(ALPHA));
+  }
+
+  @Test
+  public void computeConfidenceInterval_callsNoiseComputeConfidenceIntervalCorrectly() {
```
MashaTelyatnikova, line 396: callsNoiseAndComputeConfidenceIntervalCorrectly

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -372,6 +386,157 @@ public void computeResult_resultAlwaysInsideProvidedBoundaries() {
...
+    int maxPartitionsContributed = 1;
+    int maxContributionsPerPartition = 3;
+    double alpha = 0.5;
+    mean =
```
MashaTelyatnikova, line 400: can't we use default mean for that test? the one which was created during the set up?

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
```diff
@@ -372,6 +386,157 @@ public void computeResult_resultAlwaysInsideProvidedBoundaries() {
...
+
+  @Test
+  public void computeConfidenceInterval_positiveSumUpperBound() {
+    // Sum confidence interval.
+    mockDoubleConfInt(ConfidenceInterval.create(0, 5));
+    // Count confidence interval.
+    mockLongConfInt(ConfidenceInterval.create(2, 5));
+    mean.computeResult();
+
+    // mean_upperbound = sum_upperBound / count_lowerBound + midPoint =  5.0 / 2.0 + (1.0+9.0)/2.0 = 7.5
```
MashaTelyatnikova, line 445: can you make spaces in the statement consistent? and in other tests as well? (below)

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/BoundedMean.java
kaelhawaty: It needs to be package private for testing. I also think it is a useful function to have it on package level.

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
kaelhawaty: This is not something I added, It was already in tests. I merely changed it to remove the noise parameter. 

Aside from that, I feel from readability point if you are using mockDoubleNoise(...) with a different value than default, you should also specify in the same test what mockLongNoise(..) value should be even if it is the same. You won't have to look back up in the setup method to know what the default value is but that is just my personal opinion.

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
kaelhawaty: It should be "calls**Noise's**ComputeConfidenceIntervalCorrectly". The suggested name seems to imply that it callsNoise and actually computes confidence interval correctly or that it verifies addNoise(..) function call and computeConfidenceInterval(..) function call.

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
kaelhawaty: https://github.com/google/differential-privacy/blob/b3e72b7593703fbc77fc29bdb2c6b7b73e5067fc/java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java#L193

Even though it is equal to default mean it is still there in the code in computeResult_callsNoiseCorrectly().
Also if I am testing it is passing parameters correctly, isn't better to clarify in the same block what parameters I am using?

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
MashaTelyatnikova: not sure, we can say the same when specifying , e.f, upper and lower bounds but leave other params out of context, because only these value matter in this tests, everything else is not improtant.

MashaTelyatnikova: but you can keep it since it was before, we can remove it later if needed

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
MashaTelyatnikova: Oh, I am sorry, I missed that on the line 417 you call computeConfidenceInterval and not addNoise.
Maybe just callsNoiseCorrectly?

since when we verify addNoise in other tests we don't specify what exactly it calls, there are many other methods.

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
MashaTelyatnikova: I don't think this example shows that we just want to list all used params in the tests, they used new one because new parameters are interesting in terms that sensitivities were calculated and passed correctly, etc.

If in your test parameters which you used are also "more interesting" that in the default version then you can use it as well. If not (which means with default parameters we can show as well that sensitivities, etc were calculated correclty) then better to use  the default one.

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/BoundedMean.java
MashaTelyatnikova: why fo you think it is useful outside of the scope of tests)? We have our public method, this method represent technicality of how we calculate these intervals, and shouldn't be important for other "users" (clients of the library or the library itself) because this algorithm might change in the future (but the public API will remain the same).

So since we need it for the test mark it please @VisibleForTesting

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/BoundedMean.java
kaelhawaty: As we discussed offline, We agreed to keep it package private with @**VisibleForTesting** annotation since it is used in testing. There is also a certain use case that might be reasonable for Variance yet not the main reason it is package private.

java/tests/com/google/privacy/differentialprivacy/BoundedMeanTest.java
kaelhawaty: As we discussed offline, We also agreed to keep it with non-default values (e.g non-equal maxPartitionsContributed, maxContributionsPerPartition since they are equal in the default mean) to also detect that sensitivities were calculated and passed correctly which proved to be useful in #53

MashaTelyatnikova (approved):

MashaTelyatnikova (approved):

PR #57, opened 2020-10-02
Golang/Java fix: Confidence interval functions cannot be called after serialization

kaelhawaty:


go/dpagg/count.go
```diff
@@ -194,6 +195,9 @@ func (c *Count) ThresholdedResult(thresholdDelta float64) *int64 {
 //
 // See https://github.com/google/differential-privacy/tree/main/common_docs/confidence_intervals.md.
 func (c *Count) ComputeConfidenceInterval(alpha float64) (noise.ConfidenceInterval, error) {
+	if c.isSerialized {
```
MashaTelyatnikova, line 198: I think here we should only check resultReturned (if it was serialized then it automatically means that there was no result returned).

Also I think we need more general flag: isUsable, isUsed, smth like that (I don't have a strong opinion about the name).

Why:

we also have merge method. It sets resultReturned to true as well, it will break confidence intervals.
We could either add third flag (isMerge), or make one general flag which signals whether this instance can be used in computations or not.

And we need to check this flag (along with resultReturned) in Result(), Merge() [e,g see checkMergeCount] and Serialize/encode.

The same for Java.
And we need to consider all possible combinations of calls of methods in tests, to cover as much as possible.

MashaTelyatnikova (changes requested):

go/dpagg/count_test.go
```diff
@@ -184,19 +215,18 @@ func TestCountMerge(t *testing.T) {
 	if got != want {
 		t.Errorf("Merge: when merging 2 instances of Count got %d, want %d", got, want)
 	}
-	if !c2.resultReturned {
-		t.Errorf("Merge: when merging 2 instances of Count for c2.resultReturned got false, want true")
+	if c2.state != Merged {
+		t.Errorf("Merge: when merging 2 instances of Count for c2.state got %v, want Merged", c2.state)
```
miracvbasaran, line 219: c2.state.String() here as well

go/dpagg/count_test.go
```diff
@@ -562,8 +642,7 @@ func TestCountEquallyInitialized(t *testing.T) {
 				l0Sensitivity:   1,
 				lInfSensitivity: 1,
 				noiseKind:       noise.LaplaceNoise,
-				count:           0,
-				resultReturned:  false},
```
miracvbasaran, line 566: We should probably check state == DEFAULT here?

go/dpagg/count_test.go
```diff
@@ -304,22 +302,83 @@ func TestCountCheckMerge(t *testing.T) {
...
+
+		err := checkMergeCount(c1, c2)
+		if (err != nil) != tc.wantErr {
+			t.Errorf("CheckMerge: No error was returned, expected error")
+		}
+		if tc.wantErr && err.Error() != tc.errMessage {
+			t.Errorf("CheckMerge: Error messages are not equal:\n got: %v\n want: %v", err.Error(), tc.errMessage)
 		}
 	}
 }
```
miracvbasaran, line 384: We could probably fit these test as test cases in CheckMergeCompatibility. Instead of calling Merge() or GobEncode, we could just modify the state of the Count object.

miracvbasaran (changes requested):

go/dpagg/mean_test.go
```diff
@@ -53,7 +53,6 @@ func TestNewBoundedMeanFloat64(t *testing.T) {
 					lInfSensitivity: 2,
 					noise:           noNoise{},
 					count:           0,
-					resultReturned:  false,
```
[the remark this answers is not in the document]
kaelhawaty: Yes I know but the tests won't run, code won't compile without these line changes. Anyway, we are planning to fix all aggregation in a single PR, right?

go/dpagg/count_test.go
```diff
@@ -131,13 +132,43 @@ func TestCountSerialization(t *testing.T) {
 		if !cmp.Equal(cUnchanged, cUnmarshalled, cmp.Comparer(compareCount)) {
 			t.Errorf("decode(encode(_)): when %s got %v, want %v", tc.desc, cUnmarshalled, c)
 		}
-		// Check that the original Count has its resultReturned set to true after serialization.
-		if !c.resultReturned {
-			t.Errorf("Count %v should have its resultReturned set to true after being serialized", c)
+		if c.state != Serialized {
+			t.Errorf("Count should have its state set to Serialized, got %v, want Serialized", c.state)
```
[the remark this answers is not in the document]
kaelhawaty: Shouldn't Go automatically call String() method when printing it out with "%v"?  The state object implements "to string conversion" method.

go/dpagg/count_test.go
kaelhawaty: Same as previous -> [link](https://github.com/google/differential-privacy/pull/57#discussion_r500838282)

go/dpagg/count_test.go
kaelhawaty: I am not sure how you can fit all these different tests into one function. Also, shouldn't we call the functions and let it modify the state of the object to check transitions rather than we change it instead?

go/dpagg/count_test.go
```diff
@@ -131,13 +132,43 @@ func TestCountSerialization(t *testing.T) {
...
 		}
 	}
 }
 
+// Tests that serialization cannot be called after Result().
+func TestCountSerializationAfterResult(t *testing.T) {
+	c := getNoiselessCount()
+	c.Result()
+
+	expectedErrMessage := "Count object cannot be serialized. Reason: Noised result was already computed and returned."
```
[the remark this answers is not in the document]
kaelhawaty: Yeah I see, the reason I did this was since we return an error message with reasoning now I thought it is a good idea to check that in every state an error message is being returned correctly. Moreover, this checks that the state transition are correct also. I agree I will remove it and just check the transition after because it is package private in Go but in Java I would just check that it throws an exception. Thanks!

go/dpagg/count_test.go
kaelhawaty: Merged them into CheckMergeStateChecks same with Serialization, This way I can use noiseless count instead of specifying count options on every test in CheckMergeCompatibility

go/dpagg/aggregationState.go
```diff
@@ -0,0 +1,40 @@
...
+//
+
+package dpagg
+
+type aggregationState int
+
+var errorMessages = []string{"", " object has been already merged.", " object has been already serialized.", "Noised result is already computed and returned."}
+var stateName = []string{"Default", "Merged", "Serialized", "ResultReturned"}
+
+const (
```
MashaTelyatnikova, line 24: As agreed change to 
const (
	Default = iota
	Merged
	Serialized
	ResultReturned
)

var errorMessages = map[int]string {
	Default: "",
	Merged: "object has been already merged",
	Serialized: "object has been already serialized",
	ResultReturned: "noised result is already computed and returned",
}

so this is not that fragil

go/dpagg/aggregationState.go
```diff
@@ -0,0 +1,40 @@
...
+	Default = iota
+	Merged
+	Serialized
+	ResultReturned
+)
+
+func (s aggregationState) errorMessage(label string) string {
+	if s == Merged || s == Serialized {
```
MashaTelyatnikova, line 32: we shouldn't have some logic here, lets use label + errMsg for every state

java/main/com/google/privacy/differentialprivacy/Count.java
```diff
@@ -267,6 +267,21 @@ private void checkMergeParametersAreEqual(CountSummary otherCount) {
         params.maxContributionsPerPartition(), otherCount.getMaxContributionsPerPartition());
   }
 
+  private enum State {
+    DEFAULT(""),
+    SERIALIZED("Count object has been already serialized."),
+    RESULTRETURNED("Noised result is already computed and returned.");
```
MashaTelyatnikova, line 273: RESULD_RETURNED

java/main/com/google/privacy/differentialprivacy/Count.java
```diff
@@ -267,6 +267,21 @@ private void checkMergeParametersAreEqual(CountSummary otherCount) {
         params.maxContributionsPerPartition(), otherCount.getMaxContributionsPerPartition());
   }
 
+  private enum State {
```
MashaTelyatnikova, line 270: make it general for all aggregations

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/AggregationState.java
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package com.google.privacy.differentialprivacy;
+
+public enum AggregationState {
+    DEFAULT(""),
+    SERIALIZED(" object has been already serialized."),
+    RESULT_RETURNED("'s noised result is already computed and returned.");
```
MashaTelyatnikova, line 22: I don't think we want this specific to error message constructions here.
Remove space before object in SERIALIZED and 
"'s " here, and make changes in the code.  I don't think we need to specify in Java wchich object's noise it was, since the exception is thrown and the caller class (the aggregation type) will appear in the logs anyway.

MashaTelyatnikova (changes requested):

java/main/com/google/privacy/differentialprivacy/AggregationState.java
kaelhawaty: Change it in Go too to have the class name in the error message of the class rather than the state.

go/dpagg/count_test.go
```diff
@@ -138,34 +137,24 @@ func TestCountSerialization(t *testing.T) {
...
-	expectedErrMessage := "Count object cannot be serialized. Reason: Count object has been already merged."
-	_, err := sourceCount.GobEncode()
-	if err == nil {
-		t.Errorf("GobEncode: No error was returned, expected error")
-	}
-	if err.Error() != expectedErrMessage {
-		t.Errorf("GobEncode: Error messages are not equal:\n got: %v\n want: %v", err.Error(), expectedErrMessage)
+		_, err := c.GobEncode()
+		if (err != nil) != tc.wantErr {
+			t.Errorf("GobEncode: for err got %v, want %t", err, tc.wantErr)
```
miracvbasaran, line 156: We should print the state here (Same for other tests)

go/dpagg/dpagg_test.go
```diff
@@ -55,12 +55,22 @@ func (noNoise) Threshold(_ int64, _, _, _, _ float64) float64 {
 	return 5
 }
 
+// If noNoise is not initialized with a noise distribution, confidence interval functions will return a default confidence interval i.e [0,0].
```
miracvbasaran, line 58: nit: "interval, i.e."

miracvbasaran (changes requested):

kaelhawaty:
Rebased on main branch to get new changes, added states to all other aggregation in Java/Go

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
```diff
@@ -25,6 +25,8 @@
 import com.google.differentialprivacy.Data.ValueType;
 import com.google.differentialprivacy.SummaryOuterClass.BoundedSumSummary;
 import com.google.protobuf.InvalidProtocolBufferException;
+
+import java.io.Serializable;
```
MashaTelyatnikova, line 29: why it was added?

go/dpagg/sum_test.go
```diff
@@ -149,9 +168,28 @@ func TestBoundedSumFloat64Serialization(t *testing.T) {
...
+		{Default, false},
+		{Merged, true},
+		{Serialized, true},
+		{ResultReturned, true},
+	} {
+		bs := getNoiselessBSF()
+		bs.state = tc.state
+
+		if _, err := bs.GobEncode(); (err != nil) != tc.wantErr {
+			t.Errorf("GobEncode: when state %v for err got %v, want %t", tc.state, err, tc.wantErr)
```
miracvbasaran, line 192: nit: "wantErr" instead of "want" inside the string. (Could you also check the other aggregations for the same fix?)

go/dpagg/sum_test.go
```diff
@@ -815,26 +779,48 @@ func TestCheckMergeBoundedSumFloat64(t *testing.T) {
...
+		{Merged, Default, true},
+	} {
+		bs1 := getNoiselessBSF()
+		bs2 := getNoiselessBSF()
+
+		bs1.state = tc.state1
+		bs2.state = tc.state2
+
+		if err := checkMergeBoundedSumFloat64(bs1, bs2); (err != nil) != tc.wantErr {
+			t.Errorf("CheckMerge: when states [%v, %v] for err got %v, want %t", tc.state1, tc.state2, err, tc.wantErr)
```
miracvbasaran, line 819: nit: Again, "wantErr" instead of "want",

MashaTelyatnikova (changes requested):

go/dpagg/sum_test.go
```diff
@@ -1083,21 +1087,43 @@ func TestSumComputeConfidenceIntervalForFloat64Computation(t *testing.T) {
...
+		{ResultReturned, false},
+		{Default, true},
+		{Merged, true},
+		{Serialized, true},
+	} {
+		bs := getNoiselessBSI()
+		bs.state = tc.state
+
+		if _, err := bs.ComputeConfidenceInterval(0.1); (err != nil) != tc.wantErr {
+			t.Errorf("ComputeConfidenceInterval: when state %v for err got %v, want %t", tc.state, err, tc.wantErr)
```
miracvbasaran, line 1105: "wantErr"

miracvbasaran (changes requested):

go/dpagg/sum_test.go
kaelhawaty: I have added that. I have also added it to the old merge compatibility tests since they had the same message.
https://github.com/google/differential-privacy/pull/57/files#diff-76b5b3637fcd413f3c05f48243f09d20R298-R299

go/dpagg/count.go
```diff
@@ -248,7 +250,6 @@ func (c *Count) GobDecode(data []byte) error {
 		noiseKind:       enc.NoiseKind,
 		noise:           noise.ToNoise(enc.NoiseKind),
 		count:           enc.Count,
-		resultReturned:  enc.ResultReturned,
```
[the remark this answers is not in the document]
kaelhawaty: Added explicit state set to default for each aggregation

java/main/com/google/privacy/differentialprivacy/BoundedSum.java
kaelhawaty: My bad, removed. Thanks for spotting it!

PR #65, opened 2020-12-10
Enable Gaussian Mechanism for Count

RinoOo:
-Added delta as private field to count
-Used a default value of delta for stochastic tester

### DIFF
--- a/cc/algorithms/algorithm.h
+++ b/cc/algorithms/algorithm.h
@@ -52,13 +52,15 @@ template <typename T>
 class Algorithm {
  public:
   //
-  // Epsilon is a standard parameter of differentially private
+  // Epsilon, delta are standard parameters of differentially private
   // algorithms. See "The Algorithmic Foundations of Differential Privacy" p17.
-  explicit Algorithm(double epsilon)
-      : epsilon_(epsilon), privacy_budget_(kFullPrivacyBudget) {
+  explicit Algorithm(double epsilon, double delta)
+        : epsilon_(epsilon), privacy_budget_(kFullPrivacyBudget) {
     DCHECK_NE(epsilon, std::numeric_limits<double>::infinity());
     DCHECK_GT(epsilon, 0.0);
   }
+  explicit Algorithm(double epsilon)
+      : Algorithm(epsilon, 0) {}
 
   virtual ~Algorithm() = default;
 
@@ -165,6 +167,8 @@ class Algorithm {
 
   virtual double GetEpsilon() const { return epsilon_; }
 
+  virtual double GetDelta() const { return delta_; }
+
  protected:
   // Returns the result of the algorithm when run on all the input that has been
   // provided via AddEntr[y|ies] since the last call to Reset.
@@ -180,6 +184,7 @@ class Algorithm {
   static constexpr double kFullPrivacyBudget = 1.0;
 
   const double epsilon_;
+  const double delta_;
   double privacy_budget_;
 };
 
--- a/cc/algorithms/algorithm.h
+++ b/cc/algorithms/algorithm.h
@@ -55,12 +55,11 @@ class Algorithm {
   // Epsilon, delta are standard parameters of differentially private
   // algorithms. See "The Algorithmic Foundations of Differential Privacy" p17.
   explicit Algorithm(double epsilon, double delta)
-        : epsilon_(epsilon), privacy_budget_(kFullPrivacyBudget) {
+      : epsilon_(epsilon), privacy_budget_(kFullPrivacyBudget) {
     DCHECK_NE(epsilon, std::numeric_limits<double>::infinity());
     DCHECK_GT(epsilon, 0.0);
   }
-  explicit Algorithm(double epsilon)
-      : Algorithm(epsilon, 0) {}
+  explicit Algorithm(double epsilon) : Algorithm(epsilon, 0) {}
 
   virtual ~Algorithm() = default;
 
--- a/cc/algorithms/algorithm.h
+++ b/cc/algorithms/algorithm.h
@@ -55,7 +55,7 @@ class Algorithm {
   // Epsilon, delta are standard parameters of differentially private
   // algorithms. See "The Algorithmic Foundations of Differential Privacy" p17.
   explicit Algorithm(double epsilon, double delta)
-      : epsilon_(epsilon), privacy_budget_(kFullPrivacyBudget) {
+      : epsilon_(epsilon), delta_(delta), privacy_budget_(kFullPrivacyBudget) {
     DCHECK_NE(epsilon, std::numeric_limits<double>::infinity());
     DCHECK_GT(epsilon, 0.0);
   }
--- a/cc/algorithms/count.h
+++ b/cc/algorithms/count.h
@@ -105,8 +105,11 @@ class Count : public Algorithm<T> {
   uint64_t GetCount() const { return count_; }
 
   // The constructor and count_ are non-private for testing.
-  Count(double epsilon, double delta, std::unique_ptr<NumericalMechanism> mechanism)
-      : Algorithm<T>(epsilon, delta), count_(0), mechanism_(std::move(mechanism)) {}
+  Count(double epsilon, double delta,
+        std::unique_ptr<NumericalMechanism> mechanism)
+      : Algorithm<T>(epsilon, delta),
+        count_(0),
+        mechanism_(std::move(mechanism)) {}
 
  private:
   void AddMultipleEntries(const T& v, uint64_t num_of_entries) {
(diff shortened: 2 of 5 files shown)


cc/algorithms/count.h
```diff
@@ -129,6 +130,7 @@ class Count<T>::Builder
     ASSIGN_OR_RETURN(mechanism, AlgorithmBuilder::UpdateAndBuildMechanism());
 
     return absl::WrapUnique(new Count<T>(AlgorithmBuilder::GetEpsilon().value(),
+                                         AlgorithmBuilder::GetDelta().value_or(1e-5),
```
dibakch, line 133: I think the default value for delta should be 0.  This is still backwards compatible as the default mechanism is Laplace, which does not require any delta at all.

cc/testing/stochastic_tester_test.cc
```diff
@@ -131,7 +131,7 @@ template <typename T>
 class CountNoDpError : public Count<T> {
  public:
   explicit CountNoDpError(double epsilon)
-      : Count<T>(epsilon, LaplaceMechanism::Builder()
+      : Count<T>(epsilon, 1e-5, LaplaceMechanism::Builder()
```
dibakch, line 134: Same here, this should also work with delta = 0

cc/algorithms/count.h
```diff
@@ -115,6 +115,7 @@ class Count : public Algorithm<T> {
 
   uint64_t count_;
   std::unique_ptr<NumericalMechanism> mechanism_;
+  double delta_;
```
dibakch, line 118: I'm wondering if we should add `delta_` to `Algorithm`.  We can also add a second constructor where we can set both `epsilon_`, and `delta_`.  WDYT?

dibakch:
Could you add tests for count that use a `GaussianMechanism`?  We should have at least one for the 'normal' case and one where it fails when no delta is set.

dibakch (commented):
Could you add tests for count that use a `GaussianMechanism`? We should have at least one for the 'normal' case and one where it fails when no delta is set.

cc/algorithms/count.h
RinoOo: Fixed in the next commit

cc/testing/stochastic_tester_test.cc
RinoOo: Fixed in the next commit.

cc/algorithms/count.h
RinoOo: Fixed in the next commit.

cc/algorithms/algorithm.h
```diff
@@ -55,7 +55,10 @@ class Algorithm {
   // Epsilon is a standard parameter of differentially private
   // algorithms. See "The Algorithmic Foundations of Differential Privacy" p17.
   explicit Algorithm(double epsilon)
-      : epsilon_(epsilon), privacy_budget_(kFullPrivacyBudget) {}
+      : epsilon_(epsilon), delta_(kDefaultDelta), privacy_budget_(kFullPrivacyBudget) {}
```
dibakch, line 58: Can you call the other constructor from this constructor with delta set to 0?

cc/algorithms/algorithm.h
```diff
@@ -195,6 +201,13 @@ class AlgorithmBuilder {
                    << " is being used. Consider setting your own epsilon based "
                       "on privacy considerations.";
     }
+
+    if (!delta_.has_value()) {
```
dibakch, line 205: I think we do not need to make this distinction.  Delta will be set as 0 by default and the numerical mechanism might fail (if it requires delta).

cc/algorithms/count.h
```diff
@@ -115,6 +115,7 @@ class Count : public Algorithm<T> {
 
   uint64_t count_;
   std::unique_ptr<NumericalMechanism> mechanism_;
+  //double delta_;
```
dibakch, line 118: Can you remove this line?

cc/algorithms/algorithm.h
RinoOo: Fixed in the next commit.

cc/algorithms/algorithm.h
RinoOo: Fixed in next commit.

cc/algorithms/algorithm.h
```diff
@@ -55,7 +55,10 @@ class Algorithm {
   // Epsilon is a standard parameter of differentially private
   // algorithms. See "The Algorithmic Foundations of Differential Privacy" p17.
   explicit Algorithm(double epsilon)
```
dibakch, line 57: nit:  Move this below the other constructor as this calls the other constructor

cc/algorithms/algorithm.h
```diff
@@ -177,6 +182,7 @@ class Algorithm {
   static constexpr double kFullPrivacyBudget = 1.0;
 
   const double epsilon_;
+  const double delta_;
```
dibakch, line 185: Where is `delta_`/`GetDelta` used?

cc/algorithms/count_test.cc
```diff
@@ -192,5 +194,15 @@ TEST(CountTest, MemoryUsed) {
   EXPECT_GT((*count)->MemoryUsed(), 0);
 }
 
+TEST(CountTest, DeltaNotSetGaussian) {
```
dibakch, line 197: Can you also add a test where building the algorithm succeeds?

cc/algorithms/algorithm.h
RinoOo: We use delta_ for the constructor. Also I think we need getDelta in the future for budget splitting and bounded algorithms

cc/algorithms/algorithm.h
```diff
@@ -54,8 +54,11 @@ class Algorithm {
   //
```
Kulankhina, line 54: nit: Does this comment refer to epsilon in the constructor or epsilon as a class member? If it's the former, then I guess we should mention delta here too.

cc/algorithms/count_test.cc
```diff
@@ -192,5 +194,29 @@ TEST(CountTest, MemoryUsed) {
...
+  HasSubstr("Delta has to be set")));
+}
+
+TEST(CountTest, BasicGaussian) {
+  std::vector<int> c = {1, 2, 3, 4, 2, 3};
+  auto count =
+      typename Count<int>::Builder()
+          .SetEpsilon(10)
+          .SetDelta(1e-5)
+          .SetLaplaceMechanism(absl::make_unique<differential_privacy::GaussianMechanism::Builder>())
```
Kulankhina, line 213: Just double-checking something. "SetLaplaceMechanism" is the deprecated method IIRC.
There is no "SetMechanism" method yet, right? If it has already been added, then I believe we should use it here and in the other test.

Kulankhina (approved):
The changes LGTM overall. I added a couple of minor optional comments. Thanks!

cc/algorithms/count_test.cc
RinoOo: I already have a branch for that, But Chris and I decided to push it later.

cc/algorithms/algorithm.h
dibakch: Good point!  I think we should mention Definition 2.4 of this book (it is on page 17 and is the epsilon-delta definition).

cc/algorithms/count_test.cc
dibakch: We can deprecate it later when we found a good solution for choosing the noise in the builder.

cc/algorithms/algorithm.h
dibakch: Yes, this makes sense.  Typically we would only add code when it is required.  But since this will be required for the next PR when adding Gaussian to BoundedSum, we can keep it as is.

cc/algorithms/algorithm.h
RinoOo: Fixed in the next commit

dibakch:
I'm currently adding `clang-format` configuration to the repository.  However, for this change, could you please add a file named `.clang-format` in the root directory of the repository and add the following contents:
```
---
Language: Cpp
BasedOnStyle: Google
SortIncludes: false
...

```
You can then run `clang-format -i *` in the `algorithm` directory.

Thanks :)

dibakch:
I got a merge conflict internally.  I pushed the internal version to GitHub.  Could you try to merge this?

cc/algorithms/count_test.cc
```diff
@@ -43,11 +43,11 @@ class CountTestPeer {
 
 namespace {
 
-using ::differential_privacy::test_utils::ZeroNoiseMechanism;
```
dibakch, line 46: Could you undo sorting of include files here?

RinoOo: Fixed in the next commit.

cc/algorithms/count_test.cc
```diff
@@ -209,5 +209,32 @@ TEST(CountTest, MemoryUsed) {
...
+TEST(CountTest, DeltaNotSetGaussian) {
+  auto failed_count =
+      Count<double>::Builder()
+          .SetEpsilon(0.5)
+          .SetLaplaceMechanism(
+              absl::make_unique<
+                  differential_privacy::GaussianMechanism::Builder>())
+          .Build();
+  EXPECT_THAT(failed_count, StatusIs(absl::StatusCode::kInvalidArgument,
+                                     HasSubstr("Delta has to be set")));
```
dibakch, line 221: The string should be "Delta must be set"

dibakch:
Thanks a lot Simin! :)

PR #84, opened 2021-07-27
Fix a rare privacy bug in DistinctPerKey in Privacy on Beam.

miracvbasaran:
The bug occurred when there are outlier users in the input that
contribute to many partitions or to many values AND the values
contributed are the same as values from other users (the second part
is critical, if the contributed values only come from a single user then
the bug does not occur). Then, the output might not have be DP due to
incorrect contribution bounding. See the comments in the newly added
tests for concrete examples of when/how the bug used to occur.

This is cherry-picked from the main branch commit
e149618d032f97b476fecc5839a278cc680def08.

This commit includes a minor change compared to the one on the main
branch: It removes the error output for kv.Codec.Encode/Decode
functions used in aggregations.go since the error output for these
functions weren't implemented in v1.0.0.


dibakch (approved):
LGTM!  I've just checkout the branch locally and building and testing was successful.

gifhuppp:
e149618d032f97b476fecc5839a278cc680def08.

PR #87, opened 2021-10-27
C++: Fix multiple contributions for implicit ApproxBounds

dibakch:
This is a backport of the fix in the main branch
ae9c6b6d5b7e772837ae336d1b3092683481ec16

This change contains a critical fix for cases where ApproxBounds is used
implicitly (not specified in the builder) and we use multiple
contributions per user or multiple contributions per partition for sum,
mean, variance, and stddev.


Piyaphartoi:
@dibakch

PR #99, opened 2022-01-27
Golang: fix confidence interval calculations with the correct confidence level

osuketh:
Fixed confidence interval for the numerator because it was calculated using the confidence level for the denominator in Golang's mean implementation. The same was true for the confidence interval for the denominator.

This corresponds to the Java code [here](https://github.com/google/differential-privacy/blob/d51eedc0beea536ea563ade62116f37633eebf4d/java/main/com/google/privacy/differentialprivacy/BoundedMean.java#L230-L231).


miracvbasaran:
Hey @osuketh,

Thanks for the PR, good catch!

It seems that using alphaNum for the denominator and alphaDen for the numerator doesn't change the output significantly since in both versions (this PR and the one currently in the repository) we try 1000 different value pairs for alphaNum and alphaDen and pick the best one. The outputs are only very slightly different (perhaps by a margin<1e-4 for lower & upper bounds) because the set of the value pairs are somewhat different. Both versions would give a "correct" confidence interval I guess, that's why the tests didn't fail, but using alphaNum for the numerator and alphaDen for the denominator as intended is obviously a good idea.

PR #123, opened 2022-08-02
Added operator overloads for PLD composition.

lumip:
Suggestion:
Simple composition of `pld1`, `pld2` can now be expressed using `pld1 + pld2`,
`k`-fold self-compositition as `k * pld1`. This allows for easy-to-read
construction of complex PLD compositions, e.g.,
`pld = k12 * (k1 * pld1 + pld2) + k3 * pld3`
instead of
`pld = pld1.self_compose(k1).compose(pld2).self_compose(k12).compose(pld3.self_compose(k3))`.

### DIFF
--- a/python/dp_accounting/privacy_loss_distribution.py
+++ b/python/dp_accounting/privacy_loss_distribution.py
@@ -657,6 +657,14 @@ class PrivacyLossDistribution:
     pmf_add = self._pmf_add.self_compose(num_times, tail_mass_truncation)
     return PrivacyLossDistribution(pmf_remove, pmf_add)
 
+  def __add__(self, other: 'PrivacyLossDistribution') -> 'PrivacyLossDistribution':
+    return self.compose(other)
+
+  def __mul__(self, num_times: int) -> 'PrivacyLossDistribution':
+    return self.self_compose(num_times)
+
+  __rmul__ = __mul__
+
 
 def identity(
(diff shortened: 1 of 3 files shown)


pritkamath:
Thanks a lot for the suggestion!
The suggested change is indeed elegant, with `+` denoting convolution as a group operation, and can potentially simplify some complex use cases (e.g. it enables composing an array of PLDs as simply `sum(pld_array)`).

However, after a lot of deliberation, we decided to not include this in the library, due to the following reasons:
1. Due to truncation, the composition operation is not necessarily associative, namely `pld1 + (pld2 + pld3)` may not be same as `(pld1 + pld2) + pld3`, and moreover the running time might also depend on the order of compositions. So leaving it to the mercy of the `+` operator might hide some under-the-hood information from the user.

2. Similarly, `pld + pld + ... + pld` (`k` times) may not exactly equal `k * pld`, since truncation is performed differently.

3. In the long run, our goal is that non-privacy-expert users of the library (who just want to compute privacy parameters or noise parameters) should use `DpEvent` and `RdpAccountant`/`PLDAccountant` classes, which removes the need to explicitly use the `.compose` and `.self_compose` methods in the `pld` library. So we only foresee this operator overloading as helpful for differential-privacy experts, which we feel is a narrow use case to justify ignoring the issues above.

However, we would love to hear more about your particular use case that motivated this change. If it is a general enough and compelling use case, we could reconsider our decision.

Thank you again for your contribution!

PR #129, opened 2022-10-08
Upgrade the rules_jvm_ external dependency to 4.4.2

jameskyle:
## Why upgrade `rules_jvm_external`?

The current version of `jvm_export` supports uploading to S3 hosted repositories. I needed this for my current project.

## Why reorganize the WORKSPACE.bazel file?

For some reason, if the differntial_privacy_deps.bzl file is loaded later on in the workspace changes never take effect. I assume there's some transitive dependency that overrides if it's not called first.

## Why create a gitignore?

House cleaning? :)

### DIFF
--- a/java/.gitignore
+++ b/java/.gitignore
@@ -1,0 +1,4 @@
+bazel-bin
+bazel-java
+bazel-out
+bazel-testlogs
--- a/java/WORKSPACE.bazel
+++ b/java/WORKSPACE.bazel
@@ -20,15 +20,6 @@ local_repository(
     name = "com_google_differential_privacy",
     path = "..",
 )
-
-# Load dependencies for the base workspace.
-load("@com_google_differential_privacy//:differential_privacy_deps.bzl", "differential_privacy_deps")
-differential_privacy_deps()
-
-# Protobuf transitive dependencies.
-load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
-protobuf_deps()
-
 # Load maven rules.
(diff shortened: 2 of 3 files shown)


jameskyle:
Anything else I need to do for this?

java/.gitignore
```diff
@@ -0,0 +1,4 @@
+bazel-bin
```
dibakch, line 1: This is a great idea!

How about adding a `.gitignore` on the top level of the repo, so that we do not track those folders in all nested workspaces?

java/WORKSPACE.bazel
```diff
@@ -42,6 +33,14 @@ rules_jvm_external_deps()
 load("@rules_jvm_external//:setup.bzl", "rules_jvm_external_setup")
 rules_jvm_external_setup()
 
+# Load dependencies for the base workspace.
```
dibakch, line 36: Can we move the
```
load("@com_google_java_differential_privacy//:dp_java_deps.bzl", "dp_java_deps")
dp_java_deps()
```
Also just before this line?  This way, we would follow the instructions as in https://github.com/bazelbuild/rules_jvm_external/releases/tag/4.4.2.

dibakch (commented):
Thanks for your PR and your patience :)  I've added two comments.  Otherwise LGTM.

dibakch:
It would be great to have this in for our next release.  As this is time sensitive and we do not know how much time you have on this PR, we'll accept this PR and work on the suggested changes internally.

Thanks for your contribution!

dibakch:
Merging this to a temporary branch before merging it with main, so that we can fix open issues.

PR #268, opened 2024-04-30
Smaller fixes for DP Auditorium README links

amdjedbens:
DP Auditorium:
* Fix README links to point to correct folders.


amdjedbens:
Hello, I hope this gets approved, so users can get to the right links when using the tool.

Issue #274, opened 2024-11-04
Feature: Double Mixture of Gaussians for dp_accounting


jan-schuchardt:
Hi,

https://github.com/google/differential-privacy/commit/af761ba5bc7fb0007ed8bdde62db77046d1aff8b added "MixtureGaussianPrivacyLoss" to `dp_accounting/pld/privacy_loss_mechanism.py`, which enables privacy accounting for mechanisms that are dominated by a single Gaussian and a mixture of Gaussians.

For certain use cases, it might be nice to generalize this to mechanisms dominated by two mixture of Gaussians, each with different mixture weights and means/sensitivities.
For instance, such a "DoubleMixtureGaussianPrivacyLoss" can be used for efficient (albeit non-tight) accounting when considering substitutions of records, see Sections 6.2 & 6.3 in https://proceedings.mlr.press/v108/koskela20b.html.
It is also needed for tight accounting of group privacy under insertion/removal, see Theorem 3.8 in https://arxiv.org/abs/2403.04867.

I already had to implement this for my last project, see https://github.com/jan-schuchardt/group_amplification/blob/master/group_amplification/privacy_analysis/composition/pld/accounting.py.
So if you'd like, I could clean my code up a bit and create a pull request.

Let me know if you thank that this would be useful addition :)

arung54:
Hi Jan,

Thanks for offering to contribute! We would be interested in adding your accounting code, please go ahead and submit a PR when you feel the code is ready. 

To anticipate some feedback on that PR - it seems like both DoubleMixtureLaplacePrivacyLoss and DoubleMixtureGaussianPrivacyLoss reuse many methods from MixtureGaussianPrivacyLoss - do you think it's possible to, as part of your cleanup, inherit these methods instead (e.g. from a new abstract class that all three could inherit the shared methods from)? This will make it easier to maintain the new code on our end after the PR is accepted :)

Thanks!
Arun

jan-schuchardt:
Hi Arun,

I've created a draft pull request, see https://github.com/google/differential-privacy/pull/277.
It would be great if you could have a look :)

Best,
Jan

PR #276, opened 2024-11-13
Add DoubleMixture privacy losses

jan-schuchardt:
Hi,

this is a draft pull request related to issue https://github.com/google/differential-privacy/issues/274.

I have implemented instances of AdditiveNoisePrivacyLoss whose privacy loss distribution is dominated by a pair of two mixture distributions. The class hierarchy looks as follows:

```
DoubleMixturePrivacyLoss
├── DoubleMixtureLaplacePrivacyLoss
├── DoubleMixtureGaussianPrivacyLoss
│   ├── MixtureGaussianPrivacyLoss
```

I have kept the existing `MixtureGaussianPrivacyLoss` as a special case of `DoubleMixtureGaussianPrivacyLoss` for backwards compatibility and because it has certain optimizations that are not implemented in its super classes.

__Specifically, the `DoubleMixtureXYZ` classes are less optimized in the following sense:__
* No caching of constant terms in the privacy loss
* No specialized heuristic for finding binary search bounds

These optimization cannot be trivially generalized to the super classes, because they rely on the (inverse) privacy loss decomposing into a sum of single-distribution privacy losses. This is not the case when we have two mixtures.

Before I create a final pull request, it would be great if we could discuss the following questions:
* Can we live with the above optimization-related issues?
* The `DoubleMixtureXYZ` classes are currently only being tested via tests for `MixtureGaussianPrivacyLoss`. Do we need separate tests for each class?

### DIFF
--- a/python/dp_accounting/dp_accounting/pld/privacy_loss_distribution.py
+++ b/python/dp_accounting/dp_accounting/pld/privacy_loss_distribution.py
@@ -1330,6 +1330,138 @@ def from_discrete_gaussian_mechanism(
                                        sampling_prob)
 
 
+def from_double_mixture_gaussian_mechanism(
+    standard_deviation: float,
+    sensitivities_upper: Sequence[float],
+    sensitivities_lower: Sequence[float],
+    sampling_probs_upper: Sequence[float],
+    sampling_probs_lower: Sequence[float],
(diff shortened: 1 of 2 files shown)


jan-schuchardt:
PS: I signed

PR #283, opened 2025-01-02
Principal

rolishack:
Fix a privacy-impacting bug in ThresholdedResult() function of dpagg.Count that leads to slightly higher than intended delta. The bug is a result of converting floating-point threshold to an integer. Consider the following to see how the bug would occur:

Assume the noisy count is 37 and the threshold computed from noise parameters & threshold delta is 37.3. This means that we should not be returning the result. However, converting the threshold (37.3) to an int64 truncates the decimal part per go specification, making it 37. Threshold (37) is smaller than or equal to noisy result (37), so we return the result.

To fix this, we round the threshold up to the nearest integer before converting it to an int64.

Although this problem did not exist for dpagg.BoundedSumInt64, we modify the code there for consistency.

### DIFF
--- a/cc/algorithms/approx-bounds.h
+++ b/cc/algorithms/approx-bounds.h
@@ -388,6 +388,10 @@ class ApproxBounds : public Algorithm<T> {
     return report;
   }
 
+  // Returns a pointer to the mechanism for testing.  Does not transfer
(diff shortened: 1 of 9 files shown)


rolishack (commented):
floating

PR #404, opened 2026-04-13
Validate decoded fields in GobDecode to prevent nil noise, invalid bounds, and division by zero

AndyBurecovics:
GobDecode functions in the dpagg package do not validate deserialized fields, allowing crafted payloads to produce aggregation objects that panic or return incorrect results on Result().

This patch adds post-decode validation mirroring the checks already enforced by constructors:

- NoiseKind maps to a valid Noise implementation (prevents nil pointer dereference)
- Epsilon, delta, sensitivities pass the same range checks as New*() constructors
- Bounds satisfy lower < upper (prevents division by zero in clamping)
- TreeHeight and BranchingFactor are positive (prevents division by zero in quantile tree)

A new helper file gobdecode_checks.go centralizes the common validation logic.

### DIFF
--- a/go/dpagg/gobdecode_checks.go
+++ b/go/dpagg/gobdecode_checks.go
@@ -1,0 +1,41 @@
+package dpagg
+
+import (
+	"fmt"
+
+	"github.com/google/differential-privacy/go/v4/checks"
+	"github.com/google/differential-privacy/go/v4/noise"
+)
+
+// validateDecodedNoise rejects NoiseKind values that don't correspond to a
+// concrete Noise implementation; without this gate, Result() panics on a nil
+// receiver after deserializing a tampered payload.
+func validateDecodedNoise(kind noise.Kind) (noise.Noise, error) {
+	n := noise.ToNoise(kind)
+	if n == nil {
+		return nil, fmt.Errorf("unsupported NoiseKind value %d", kind)
+	}
+	return n, nil
+}
+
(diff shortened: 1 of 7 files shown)


go/dpagg/gobdecode_checks.go
miracvbasaran, line 1: It might make sense to move this to `go/checks/` directory.

go/dpagg/sum.go
```diff
@@ -693,6 +700,13 @@ func (bs *BoundedSumFloat64) GobDecode(data []byte) error {
 	if err != nil {
 		return fmt.Errorf("couldn't decode BoundedSumFloat64 from bytes")
 	}
+	nObj2, err := validateDecodedAggregation(enc.Epsilon, enc.Delta, enc.LInfSensitivity, enc.L0Sensitivity, enc.NoiseKind)
```
miracvbasaran, line 703: nit: `nObj2` -> `nObj`

go/dpagg/mean.go
```diff
@@ -379,6 +379,9 @@ func (bm *BoundedMean) GobDecode(data []byte) error {
 	if err != nil {
 		return fmt.Errorf("couldn't decode BoundedMean from bytes")
 	}
+	if err := checks.CheckBoundsFloat64(enc.Lower, enc.Upper); err != nil {
```
miracvbasaran, line 382: For mean & variance, I think we could call validateDecodedAggregation on the underlying Count and Sum objects for completeness.

PR #412, opened 2026-04-25
fix: add parameter validation to GobDecode for all DP aggregation types

AndyBurecovics:
## Problem

Every aggregation type in `go/dpagg` (Count, BoundedSumInt64, BoundedSumFloat64, BoundedQuantiles, PreAggSelectPartition) implements GobDecode without validating deserialized parameters. The constructors run 20+ checks from `checks/checks.go` before accepting parameters. GobDecode runs zero of them.

This opens three specific issues:

1. **Nil Noise dereference**: `noise.ToNoise()` returns nil for any NoiseKind other than 0 or 1. GobDecode feeds the deserialized NoiseKind straight into ToNoise without checking the return. An invalid NoiseKind causes a nil pointer panic when `Result()` is called.

2. **Division by zero**: BoundedQuantiles accepts `branchingFactor=0` from the payload. `getParent()` does `(index - 1) / bq.branchingFactor`, which panics on integer division by zero.

3. **Invalid sensitivity parameters**: GobDecode restores `lInfSensitivity` and bounds as independent fields without verifying consistency or basic validity (e.g. positive, finite).

Composite types (BoundedMean, BoundedVariance, BoundedStandardDeviation) are protected transitively because Gob calls GobDecode on their embedded Count and BoundedSumFloat64 fields during deserialization.

## Fix

- Validate that `noise.ToNoise(noiseKind)` returns non-nil before assignment.
- Validate `l0Sensitivity` and `lInfSensitivity` using existing check functions.
- Validate bounds ordering where applicable.
- Validate `branchingFactor >= 2` and `treeHeight` in BoundedQuantiles.
- Add tests for all validation paths.

Related: https://issuetracker.google.com/issues/501920784

### DIFF
--- a/go/dpagg/count.go
+++ b/go/dpagg/count.go
@@ -305,13 +305,23 @@ func (c *Count) GobDecode(data []byte) error {
 	if err != nil {
 		return fmt.Errorf("couldn't decode Count from bytes")
 	}
+	n := noise.ToNoise(enc.NoiseKind)
+	if n == nil {
+		return fmt.Errorf("GobDecode: invalid NoiseKind %d", enc.NoiseKind)
+	}
+	if err := checks.CheckL0Sensitivity(enc.L0Sensitivity); err != nil {
+		return fmt.Errorf("GobDecode Count: %v", err)
+	}
+	if enc.LInfSensitivity <= 0 {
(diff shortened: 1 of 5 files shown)


miracvbasaran:
Thanks for the PRs!

I'd prefer it if we finalize the existing 2 PRs first (#404, #405) before I review the next two (#411, #412). Please see my comments on the previous ones.

Cheers,

PR #414, opened 2026-04-30
Integrated the random allocation directory

moshenfeld:
Added a new pld/random_allocation package which implements the accounting algorithm proposed in "[Efficient privacy loss accounting for subsampling and random allocation](https://arxiv.org/pdf/2602.17284)".
It contains public builders for Gaussian and explicit-realization random allocation PLDs in random_allocation_api.py.
It also wires the feature into dp_accounting as RandomAllocationDpEvent in dp_event.py, and makes PLDAccountant handle Gaussian inner events in pld_privacy_accountant.py.


python/dp_accounting/dp_accounting/dp_event.py
```diff
@@ -494,3 +494,23 @@ class TruncatedSubsampledGaussianDpEvent(DpEvent):
...
   truncated_batch_size: int
   noise_multiplier: float
+
+
+@attr.s(frozen=True, slots=True, auto_attribs=True)
+class RandomAllocationDpEvent(DpEvent):
+  """Represents the random-allocation mechanism.
+
+  In this event, each element in the dataset is independently allocated to k
+  uniformly sampled calls to the mechanism out of t total calls.
```
arung54: Reference to the PLD accounting paper could be useful here? (Will also get you more visibility on the paper :) )

python/dp_accounting/dp_accounting/pld/pld_privacy_accountant.py
```diff
@@ -263,6 +264,31 @@ def _maybe_compose(self, event: dp_event.DpEvent, count: int,
           )
           self._pld = self._pld.compose(truncated_subsampled_gaussian_pld)
       return None
+    elif isinstance(event, dp_event.RandomAllocationDpEvent):
```
arung54, line 299: Is it possible to add a simple test for this event? Similar to e.g "test_gaussian_basic" in pld_privacy_accountant_test.

python/dp_accounting/dp_accounting/pld/pld_privacy_accountant.py
```diff
@@ -263,6 +264,31 @@ def _maybe_compose(self, event: dp_event.DpEvent, count: int,
...
+      ra_event: dp_event.RandomAllocationDpEvent = event
```
arung54: Is the re-naming needed here? The isinstance() should let the interpreter know there is a subevent to access?

If it's just for clarity (e.g. to avoid calling it 'event.event') maybe better to rename the field in RandomAllocationDpEvent.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+        del fft2  # free second complex buffer immediately
+    conv_full = irfft(fft1, n=fft_size, overwrite_x=True)
+    del fft1  # free complex buffer
+    conv_pmf = conv_full[:conv_full_len].copy()  # copy needed portion
+    del conv_full  # free full irfft output
+
+    # Zero negative roundoff and ghost mass outside reachable support
+    conv_pmf[conv_pmf < 0] = 0.0
+    conv_pmf[:min_idx] = 0.0
+    max_idx_plus_one = max_idx + 1
```
ryan112358: Can probably inline max_idx+1 rather than giving it a name

python/dp_accounting/dp_accounting/pld/random_allocation/ra_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+    del conv_full  # free full irfft output
+
+    # Zero negative roundoff and ghost mass outside reachable support
+    conv_pmf[conv_pmf < 0] = 0.0
+    conv_pmf[:min_idx] = 0.0
+    max_idx_plus_one = max_idx + 1
+    if max_idx_plus_one < conv_pmf.size:
+        conv_pmf[max_idx_plus_one:] = 0.0
+
+    current_finite_mass = math.fsum(map(float, conv_pmf))
```
ryan112358: I see this pattern also above, what's the benefit of this compared to np.sum(conv_pmf)?  Can you add an inline comment?

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+        prob_arr=np.asarray(prob_arr, dtype=np.float64),
+        zero_prob=float(zero_prob),
+        x_out_0=float(x_out_0),
+        log_r=float(geom_step),
+        dominates=(bound_type == BoundType.DOMINATES),
+    )
+
+
+@njit(cache=True)
+def _numba_add_single_zero_atom_cross_term(
```
ryan112358: Numba is a large dependency that I believe dp_accounting doesn't currently have. Can this be written efficiently in plain numpy? If it's needed for performance reasons we can keep it, but figured I'd ask since it looks like a pretty simple function to vectorize efficiently.

1) If pure numpy can be nearly yas efficient, prefer that.
2) If not, then prefer treating numba as an optional dependency of dp_accounting.  If it's installed, use this codepath, otherwise use a (vectorized) numpy codepath, and include a "Performance Warning" that users should install numba if using this code.  Would be good to include test coverage that the two code paths match in that case.

python/dp_accounting/dp_accounting/dp_event.py
```diff
@@ -494,3 +494,23 @@ class TruncatedSubsampledGaussianDpEvent(DpEvent):
...
+
+  Currently only GaussianDpEvent is supported as the inner event.
+
+  Attributes:
+    event: The DpEvent for each call to the mechanism (e.g., GaussianDpEvent).
+    k: Number of calls each element participates in.
```
ryan112358: For the public API it might be worth giving k and t slightly more descriptive names.  But if k/t matches the notation used in the paper, then I could see an argument for keeping it as is, so consider this optional.

python/dp_accounting/dp_accounting/dp_event_test.py
```diff
@@ -134,6 +142,12 @@ def test_to_from_named_tuple(self, event):
     assert_not_contains_named_tuples(reconstructed)
     self.assertEqual(event, reconstructed)
 
+  def test_random_allocation_event_has_no_discretization_override(self):
```
arung54: Maybe a comment on why this test is necessary would be useful?

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+
+from __future__ import annotations
+
+import math
+import os
+import warnings
+
+import numpy as np
+from dp_accounting.pld.common import compute_self_convolve_bounds
+from numba import njit
```
arung54: We may want to keep numba an optional dependency in the overall library, while still allowing users who have it installed to benefit from the njit. Can you change to something like this?

```
try:
    from numba import njit
    _HAS_NUMBA = True
except ImportError:
    _HAS_NUMBA = False

def optional_njit():
    if _HAS_NUMBA:
        return njit(cache=True)
    else:
        return lambda x: x
```

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_utils.py
```diff
@@ -0,0 +1,555 @@
+"""Utility functions for distribution operations and numerical stability."""
+
+from __future__ import annotations
+
+import math
+from typing import Any, Callable
+
+import numpy as np
+from numba import njit
```
arung54: Same comment on making numba optional applies here.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+REALIZATION_MOMENT_TOL = 1e-12
+SPACING_ATOL = 1e-12
+SPACING_RTOL = 1e-6
+MIN_GRID_SIZE = 100  # Minimum number of points in a  discretization grid.
+MAX_SAFE_EXP_ARG = math.log(np.finfo(np.float64).max)
+TAIL_SWITCH = 1e-10
+
+
+class Domain(Enum):
+    """Domain of a discrete distributsion's support."""
```
arung54: Nit: "distributsion" -> "distribution"

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+            self.prob_arr,
+            self.p_min,
+            self.p_max,
+        )
+
+        pmf_sum = math.fsum(map(float, self.prob_arr))
+        total_mass = pmf_sum + self.p_min + self.p_max
+        mass_error = abs(total_mass - 1.0)
+        if mass_error > PMF_MASS_TOL:
+            error_msg = "MASS CONSERVATION ERROR"
```
arung54: Maybe more readable if you write this all as a single assignment, e.g.

error_msg = (
        f"MASS CONSERVATION ERROR: Error={mass_error:.2e} "
        f"(tolerance={PMF_MASS_TOL:.2e}), PMF sum={pmf_sum:.15f}, "
        f"min={self.p_min:.2e}, max={self.p_max:.2e}, "
        f"Total mass={total_mass:.15f}"
    )

arung54: Rather than "MASS CONSERVATION ERROR" can we have a more descriptive statement, e.g. "PMF mass does not total 1:"

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+            error_msg += f", max={self.p_max:.2e}"
+            error_msg += f", Total mass={total_mass:.15f}"
+            raise ValueError(error_msg)
+
+        # REALS domain: both boundaries being non-zero is not allowed.
+        if self.domain == Domain.REALS and self.p_min > PMF_MASS_TOL and self.p_max > PMF_MASS_TOL:
+            raise ValueError("REALS domain: p_min and p_max cannot both be non-zero")
+
+    def truncate_edges(self, tail_truncation: float, bound_type: BoundType) -> Self:
+        """Truncate distribution edges. Computation lives in distribution_utils."""
```
arung54: I think this comment is stale and compute_truncation is in this library now?

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+        """Deep-copy this distribution while preserving representation type."""
+
+
+# =============================================================================
+# GENERAL (EXPLICIT) DISTRIBUTION
+# =============================================================================
+
+
+class SparseDiscreteDist(DiscreteDistBase):
+    """General discrete distribution with explicit support values."""
```
arung54: Can you add Attributes here

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+        super().__init__(prob_arr, p_min, p_max, domain)
+        self._validate_x_array()
+
+    def _validate_x_array(self) -> None:
+        if self._x_array.ndim != 1 or self._x_array.shape != self.prob_arr.shape:
+            raise ValueError("x and PMF must be 1-D arrays of equal length")
+        if not np.all(np.diff(self._x_array) > 0):
+            raise ValueError("x must be strictly increasing")
+
+    def get_x_array(self) -> NDArray[np.float64]:
```
arung54: Any reason to not just make x_array public? See https://google.github.io/styleguide/pyguide.html#315-getters-and-setters

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+    ) -> SparseDiscreteDist:
+        return SparseDiscreteDist(
+            x_array=self._x_array[slice(min_ind, max_ind + 1)].copy(),
+            prob_arr=new_prob_arr,
+            p_min=new_p_min,
+            p_max=new_p_max,
+            domain=self.domain,
+        )
+
+    def copy(self) -> SparseDiscreteDist:
```
arung54: Is there a reason to use this over deepcopy?

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+
+
+# =============================================================================
+# UNIFIED REGULAR-GRID DISTRIBUTION
+# =============================================================================
+
+
+class DenseDiscreteDist(DiscreteDistBase):
+    """Discrete distribution on a regular (linear or geometric) grid.
+
```
arung54: Attributes here too

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+        prob_arr: NDArray[np.float64],
+        p_min: float = 0.0,
+        p_max: float = 0.0,
+        spacing_type: SpacingType = SpacingType.LINEAR,
+        domain: Domain = Domain.REALS,
+    ) -> "DenseDiscreteDist":
+        """Create DenseDiscreteDist from x_array by extracting x_min and step."""
+        if spacing_type == SpacingType.LINEAR:
+            step = compute_bin_width(x_array)
+        else:
```
arung54: Preferable to use 
```
elif self.spacing_type == SpacingType.GEOMETRIC:
    ...
else:
    raise ValueError(f"Unknown SpacingType: {self.spacing_type}")
```

If for any reason you'd prefer not to, maybe add a comment after the else saying "self.spacing_type == SpacingType.GEOMETRIC" so it's easy to tell what the case is.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+        max_ind: int,
+    ) -> "DenseDiscreteDist":
+        if self.spacing_type == SpacingType.LINEAR:
+            new_grid = RegularGrid(
+                x_min=self.x_min + min_ind * self.step,
+                step=self.step,
+                size=new_prob_arr.size,
+                spacing_type=self.spacing_type,
+            )
+        else:
```
arung54: Same comment on else -> elif here.

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+            step=new_grid.step,
+            prob_arr=new_prob_arr,
+            p_min=new_p_min,
+            p_max=new_p_max,
+            spacing_type=self.spacing_type,
+            domain=self.domain,
+            grid=new_grid,
+        )
+
+    def copy(self) -> "DenseDiscreteDist":
```
arung54: Same question about deepcopy.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+        )
+
+
+# =============================================================================
+# PLD REALIZATION
+# =============================================================================
+
+
+class PLDRealization(DenseDiscreteDist):
+    """Linear-grid PLD realization in loss space."""
```
arung54: Add Attributes here

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+    """Compute linear spacing width for two grids and return their average."""
+    w1 = compute_bin_width(x_array_1)
+    w2 = compute_bin_width(x_array_2)
+    if not stable_isclose(a=w1, b=w2):
+        raise ValueError(f"Grid spacing must match: w1={w1:.12g} vs w2={w2:.12g}")
+    return (w1 + w2) / 2
+
+
+# =============================================================================
+# Grid Spacing Utilities
```
arung54: Our internal style is to define helper/convenience functions before their first use, ideally move these to the top of the file.

arung54: Are these used outside this library? If not consider making them private to simplify the API of this file.

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+                "Exponential moment E[exp(-L)] is infinite, not a valid PLD realization"
+            )
+        exp_moment_total = math.fsum(map(float, exp_moment_val))
+        if exp_moment_total > 1.0 + REALIZATION_MOMENT_TOL:
+            raise ValueError(
+                f"Exponential moment E[exp(-L)] = {exp_moment_total:.15f} > 1.0, "
+                "not a valid PLD realization"
+            )
+
+    def copy(self) -> "PLDRealization":
```
arung54: Same question about deepcopy

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+            # IS_DOMINATED can set p_min > 0, violating PLDRealization.p_min = 0.
+            # Delegate through a plain DenseDiscreteDist so the result is not a PLDRealization.
+            return DenseDiscreteDist(
+                x_min=self.x_min,
+                step=self.step,
+                prob_arr=self.prob_arr.copy(),
+                p_min=self.p_min,
+                p_max=self.p_max,
+            ).truncate_edges(tail_truncation, bound_type)
+        return super().truncate_edges(tail_truncation, bound_type)
```
arung54: For readability maybe put this in an elif-else or in a commented else as described above.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+    def _create_truncated(
+        self,
+        new_prob_arr: NDArray[np.float64],
+        new_p_min: float,
+        new_p_max: float,
+        min_ind: int,
+        max_ind: int,
+    ) -> "PLDRealization":
+        """Create a truncated PLD realization while preserving linear-loss semantics."""
+        del max_ind
```
arung54: To match our internal style, add a comment:
`del max_ind  # Unused.`

Optionally, you can explain in the comment why it's not used.

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+        log_terms = np.log(prob_arr[extreme_mask]) - x_vals[extreme_mask]
+        terms_extreme = np.exp(np.minimum(log_terms, MAX_SAFE_EXP_ARG))
+        terms_extreme[log_terms > MAX_SAFE_EXP_ARG] = np.inf
+        terms[extreme_mask] = terms_extreme
+
+    return terms
+
+
+# =============================================================================
+# Distribution Edge Truncation
```
arung54: Same comment on moving above methods that using it, and also maybe making the public methods private if not used outside this library.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+            sf=sf,
+            tail_truncation=PMF_min_increment,
+        )
+    else:
+        raise ValueError(f"Unknown BoundType: {bound_type}")
+
+    return bin_probs, p_left, p_right
+
+
+def _discretize_continuous_to_grid(
```
arung54: Move to above the first usage

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+    return discretize_continuous_grid(
+        dist=dist,
+        grid=grid,
+        bound_type=bound_type,
+        PMF_min_increment=tail_truncation,
+        domain=domain,
+    )
+
+
+def discretize_continuous_dist(
```
arung54: Give this or discretize_continuous_distribution a more descriptive name that clarifies the difference between them. Clarifying the difference in the docstring might also be helpful.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+    """Compute discrete PMF and boundary masses on a materialized grid."""
+    bin_probs, p_left, p_right = _compute_discrete_prob(
+        dist=dist, x_array=x_array, bound_type=bound_type, PMF_min_increment=PMF_min_increment
+    )
+
+    n = x_array.size
+    prob_arr = np.zeros(n)
+
+    if bound_type == BoundType.DOMINATES:
+        # Shift mass right: left tail (-inf, x0) -> x0,
```
arung54: Nit: Here and below, maybe x_0 instead of x0 for consistency with the next line.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+        raise ValueError(
+            f"Cannot discretize {dist_name} to a positive range, got x_min={x_array[0]}"
+        )
+
+    # 2. Map density to PMF with semantics.
+    return discretize_continuous_grid(
+        dist=dist,
+        grid=grid,
+        bound_type=bound_type,
+        PMF_min_increment=tail_truncation,
```
arung54: Here and elsewhere, name pmf_min_increment to match Google style. See https://google.github.io/styleguide/pyguide.html#316-naming

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+    spacing_type: SpacingType,
+    bound_type: BoundType,
+) -> DenseDiscreteDist:
+    """Rediscretize a distribution onto a requested grid spacing.
+
+    Remaps PMF onto a new grid with the requested spacing and discretization.
+    Implementation trims zero/tail regions, computes new grid size, then remaps
+    using domination-aware rounding (e.g., linear grids for dp_accounting output).
+
+    Algorithm 6 (`disc-dist`) in Appendix C.
```
arung54: Maybe link to the paper here directly

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions.py
```diff
@@ -0,0 +1,1390 @@
...
+    *,
+    dist: stats.rv_continuous | rv_frozen[Any, Any],
+    x_array: NDArray[np.float64],
+    bound_type: BoundType,
+    PMF_min_increment: float,
+) -> tuple[NDArray[np.float64], float, float]:
+    """Compute bin probabilities using adaptive CDF/SF increments with logcdf/logsf stability.
+
+    PMF_min_increment controls the minimum CDF/SF increment that becomes a bin mass.
+
```
arung54: Nit: Remove blank line before end of docstring.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions_test.py
```diff
@@ -0,0 +1,562 @@
...
+        grid=grid,
+        prob_arr=np.array([0.4, 0.6], dtype=np.float64),
+        domain=Domain.POSITIVES,
+    )
+
+    assert dist.grid == grid
+    assert dist.step == 1e-4
+
+
+class TestDiscritizeRange:
```
arung54: Nit: 'Discritize' -> 'Discretize.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_distributions_test.py
```diff
@@ -0,0 +1,562 @@
...
+
+import math
+
+import numpy as np
+import pytest
+from .random_allocation_distributions import (
+    DenseDiscreteDist,
+    Domain,
+    PLDRealization,
+    _compute_discrete_prob as compute_discrete_PMF,
```
arung54: Any reason to rename? If so, prefer to lowercase (see e.g. https://google.github.io/styleguide/pyguide.html#316-naming)

python/dp_accounting/dp_accounting/pld/random_allocation/ra_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+        p_min=p_min,
+        p_max=p_max,
+        domain=dist_1.domain,
+    ).truncate_edges(tail_truncation, bound_type)
+
+
+def fft_self_convolve(
+    *,
+    dist: DenseDiscreteDist,
+    T: int,
```
arung54: Here and elsewhere: The pylint checker will probably throw a complaint here, since this is a math variable matching a paper it's okay to silence this complaint, see https://google.github.io/styleguide/pyguide.html#3165-mathematical-notation

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+    *,
+    dist: DenseDiscreteDist,
+    T: int,
+    tail_truncation: float,
+    bound_type: BoundType,
+    use_direct: bool,
+) -> DenseDiscreteDist:
+    """T-fold self-convolution via FFT with optional direct exponentiation path."""
+    if not (isinstance(dist, DenseDiscreteDist) and dist.spacing_type == SpacingType.LINEAR):
+        raise TypeError("fft_self_convolve requires DenseDiscreteDist input")
```
arung54: Nit: This TypeError should probably specify 'when spacing_type == LINEAR'

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+        T=T,
+        tail_truncation=tail_truncation,
+        bound_type=bound_type,
+        convolve=fft_convolve,
+    )
+    if not (
+        isinstance(self_conv, DenseDiscreteDist) and self_conv.spacing_type == SpacingType.LINEAR
+    ):
+        raise TypeError(
+            f"Expected DenseDiscreteDist from FFT self-convolution, got {type(self_conv)}"
```
arung54: Is it possible for binary_self_convolve not to return a DenseDiscreteDist? If so, this error message should probably specify how to change inputs to this method to prevent binary_self_convolve from returning another output. If not, this line can be removed.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+def geometric_convolve(
+    *,
+    dist_1: DenseDiscreteDist,
+    dist_2: DenseDiscreteDist,
+    tail_truncation: float,
+    bound_type: BoundType,
+) -> DenseDiscreteDist:
+    """Convolve two geometric-grid distributions.
+
+    Algorithm 4 (`conv`) in Appendix C wrapper.
```
arung54: Maybe link to paper here?

python/dp_accounting/dp_accounting/pld/random_allocation/ra_convolution.py
```diff
@@ -0,0 +1,657 @@
...
+    )
+    if not (
+        isinstance(self_conv, DenseDiscreteDist) and self_conv.spacing_type == SpacingType.GEOMETRIC
+    ):
+        raise TypeError(f"Expected DenseDiscreteDist from self-convolution, got {type(self_conv)}")
+    return self_conv
+
+
+# =============================================================================
+# INTERNAL KERNEL IMPLEMENTATION
```
arung54: Move helpers above the first usage.

python/dp_accounting/dp_accounting/pld/random_allocation/ra_core.py
```diff
@@ -0,0 +1,794 @@
...
+            Must be a linear DenseDiscreteDist. ``x_min`` is rounded to the
+            nearest step multiple when forming dp_accounting's integer
+            ``lower_loss`` index.
+        pessimistic_estimate: Whether to use pessimistic estimate in dp_accounting.
+
+    Returns:
+        dp_accounting DensePLDPmf with infinity mass taken from dist.p_max.
+    """
+    if not (isinstance(dist, DenseDiscreteDist) and dist.spacing_type == SpacingType.LINEAR):
+        raise TypeError(
```
arung54: The TypeError should clarify e.g. 'and spacing_type LINEAR' as well

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_core.py
```diff
@@ -0,0 +1,794 @@
...
+def realization_remove_base_distributions(
+    *,
+    realization: PLDRealization,
+    loss_discretization: float,
+    tail_truncation: float,
+    bound_type: BoundType,
+) -> tuple[DenseDiscreteDist, DenseDiscreteDist]:
+    """Prepare remove-direction factors from a loss-space realization.
+
+    Algorithm 1 (`rand-alloc-rem`) in Appendix C.
```
arung54: Add a link to paper here

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_core.py
```diff
@@ -0,0 +1,794 @@
...
+    Returns:
+        Tuple ``(base, dual_base)`` aligned to the requested linear grid.
+
+    """
+    # Since dual can be derived only from a PLD realization, discretization can
+    # come first for DOMINATES, but dual derivation must come first for IS_DOMINATED.
+    if bound_type == BoundType.DOMINATES:
+        # Avoid inflating the grid when the target is finer than the original one.
+        effective_disc = max(realization.step, loss_discretization)
+        coarsened_base = rediscretize_dist(
```
arung54: Similar to another comment, here and elsewhere it seems like our type checker will guarantee a DenseDiscreteDist, in which case the additional type check might not be necessary (but maybe you want to check spacing anyway)

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_core.py
```diff
@@ -0,0 +1,794 @@
...
+def realization_add_base_distribution(
+    *,
+    realization: PLDRealization,
+    loss_discretization: float,
+    tail_truncation: float,
+    bound_type: BoundType,
+) -> DenseDiscreteDist:
+    """Prepare add-direction factors from a loss-space realization.
+
+    Algorithm 2 (`rand-alloc-add`) in Appendix C.
```
arung54: Add link to paper here

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    *,
+    dist: DenseDiscreteDist,
+    T: int,
+    tail_truncation: float,
+    bound_type: BoundType,
+    convolve: Callable[..., DenseDiscreteDist],
+) -> DenseDiscreteDist:
+    """Exponentiation by squaring-based self-convolution using a provided convolve function.
+
+    Algorithm 3 (`self-conv`) in Appendix C.
```
arung54: Add link to paper here

python/dp_accounting/dp_accounting/pld/random_allocation/utils.py
```diff
@@ -0,0 +1,555 @@
...
+        x_min=-(realization.x_min + realization.step * (realization.prob_arr.size - 1)),
+        step=realization.step,
+        prob_arr=dual_probs,
+        p_max=max(0.0, 1.0 - sum_prob),
+        p_min=0.0,
+    )
+
+
+# =============================================================================
+# Internal Helper Functions
```
arung54, line 240: Move these to above their first usage, including reordering them (e.g. _expand_to_grid above _align_distributions_to_union_grid)

python/dp_accounting/dp_accounting/pld/random_allocation/ra_utils.py
```diff
@@ -0,0 +1,555 @@
...
+        x_min=-(dist.x_min + dist.step * (n - 1)),
+        step=dist.step,
+        prob_arr=np.flip(dist.prob_arr),
+        p_min=dist.p_max,
+        p_max=dist.p_min,
+    )
+
+
+def calc_pld_dual(realization: PLDRealization) -> PLDRealization:
+    """Compute the paper PLD dual ``D(L)`` (Definition 3.1).
```
arung54: Link to paper here

arung54 (commented):
Hi Moshe,

No comments on the logic / implementation. Most of these comments are about style (and stuff that we will probably have to address anyway to pass the style checks) which are hopefully easy to fix. Some of these could have been prevented if I shared our style guide in advance, apologies for only pointing you to it during the review! If you haven't already, I'd also suggest fixing any issues raised by:

```
$ flake8 random_allocation/*.py
$ pyink random_allocation/*.py
$ pydocstyle --convention=google --add-ignore=D101,D102,D103,D105,D202,D402 random_allocation/*.py
$ pylint random_allocation/*.py
$ pytype random_allocation/*.py -k
```

Some top-level comments:

- Since this library is already in a folder named 'random_allocation', consider renaming some of the libraries to be shorter (e.g. from 'random_allocation_convolution' to 'convolution') for brevity.
- See https://google.github.io/styleguide/pyguide.html#22-imports - a lot of these libraries are importing methods, whereas the style guide suggests importing modules and then calling the method from the package name. Renaming to shorter module names might help with doing this while maintaining brevity.
- For the test files, can you match the style of the other test files in dp_accounting here?:
  - Use absltest instead of pytest
  - Prefer absltest.TestCase or numpy.testing assertions over assert ..., since these will give more verbose failures.
  - Put "Test" at the end of test class names rather than beginning

ryan112358:
A few minor comments

onlinedigitalservicexplacebank:
>

python/dp_accounting/dp_accounting/dp_event_test.py
moshenfeld: Removed. Instead, I renamed the variable to match dp accounting's convention

python/dp_accounting/dp_accounting/pld/pld_privacy_accountant.py
moshenfeld: No need.
Reverted.

python/dp_accounting/dp_accounting/pld/pld_privacy_accountant.py
moshenfeld: Done

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
moshenfeld: Is this implementation the right design?

python/dp_accounting/dp_accounting/pld/random_allocation/ra_core.py
moshenfeld: Done

moshenfeld:
Thank you.
I'm in the midst of going over all the comments and ensuring they are all
resolved.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_utils.py
moshenfeld: Done in the types file

python/dp_accounting/dp_accounting/dp_event.py
moshenfeld: Updated the API to num_selected (k) and num_steps (t)

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
moshenfeld: The idea was to create a joint interface for sparse and dense distributions just like in the PMF case.
I assume you didn't mean actually public but just used within the directory, so I added a _ prefix.

python/dp_accounting/dp_accounting/pld/random_allocation/ra_convolution.py
moshenfeld: It is much more stable which is critical for large arrays with tiny values as we typically use.
I added comments.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_core.py
moshenfeld: I exported it to a separate _assert_dense_linear_dist function

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
moshenfeld: Not really.
I implemented protections in a paranoid manner, where no function trust the others, but it is indeed an overkill.

python/dp_accounting/dp_accounting/pld/random_allocation/random_allocation_convolution.py
moshenfeld: The advantage is very significant.
I implemented the numpy fallback and added some comparison tests.
Does this seem right?

python/dp_accounting/dp_accounting/pld/random_allocation/ra_distributions.py
moshenfeld: I hope I didn't overdo it

python/dp_accounting/dp_accounting/pld/pld_privacy_accountant.py
```diff
@@ -23,6 +23,7 @@
 from dp_accounting import privacy_accountant
 from dp_accounting.pld import common
 from dp_accounting.pld import privacy_loss_distribution
+from dp_accounting.pld import random_allocation as _random_allocation
```
arung54: Just for consistency, I think it's fine to import without 'as _random_allocation'

arung54 (commented):
Hi Moshe,

Overall looks good. One minor comment below. In addition, I think the linter will complain if you are accessing protected methods from other classes (unless you are in a test class accessing a protected method from the tested class). I think the 'public API' methods in api.py / core.py are the main culprits for where the linter would complain.

Also, I think we will need to add a BUILD.bazel (similar to https://github.com/google/differential-privacy/blob/main/python/dp_accounting/dp_accounting/BUILD.bazel) to this folder, and update the BUILD.bazel in the parent folder. For the most part it should be pattern-matchable (also happy to help you with writing it), but since we are adding a numba dependency that may introduce some complications. Let me chat w/ other folks on our team at the right way to add numba (since right now it's not in the pip deps) and get back to you.

moshenfeld:
I think I fixed the remaining issues.

Before you check the changes, I just want to make sure I got the rules
right this time.
In the last round you had this comment
<https://github.com/google/differential-privacy/pull/414#discussion_r3229395649>
about Grid Spacing Utilities where you asked if these functions are "used
outside this library" and recommended making them private if not.
I took the "outside this library" to mean that if a method is called from a
different file inside the RA directory, it should start with an underscore,
but it seems like I've overdone it.
If I understand correctly, the right format for such functions is - remove
underscore but exclude from init. If the function is used only within the
same file, then it should start with underscore.
Did I get it this time?

Also, I hope I got the bazel style right. Please let me know if I missed
something.

Thanks
Moshe

arung54:
Sorry for the lack of clarity earlier! Yes, if a method in foo.py is only used in foo.py (and maybe foo_test.py) it should start with underscore. If it is used in bar.py in the same folder, the linter will complain if it starts with an underscore, which we need to avoid to pass style checks. And yes, if it is only used within the same folder, not including it in the init is good.

Thanks for your patience with the comments, at this point I think it's looking good generally. Let me check internally the right way to get numba added as a dependency, then I might suggest a change to one of the BUILD.bazels and then we are probably good to go to internal approval, which we can hopefully make fast. I will be OOO for the next few days due to the US holiday, but will try to check back in soon.

moshenfeld:
In this case, I think the delivered code matches your instructions.
I also checked the possibility of removing numba, which I discussed in a
separate thread.
Moshe

arung54 (approved):

arung54:
Hi Moshe,

Thanks for the updates! I think to migrate this into an internal PR we need to resolve the merge conflicts? Otherwise I will try to keep things moving on our end once that's done.

moshenfeld:
I'm either missing something or don't have the required permissions to see
them.
I can see the "This branch has conflicts that must be resolved" message but
cannot click the "resolve conflicts" button, and didn't find any other way
to see those conflicts.

arung54:
Got it, I'll ask someone on our side who's more familiar with the OSS process for how to proceed then!

ryan112358:
I'm not 100% sure, but one thing that might be happening is if the resolution of conflicts is non-trivial, you can't do it in the web interface, but you have to do it manually via the command line w/ git. I actually don't have a ton of experience with this, but I believe if you "git pull" it will sync to latest version at head and you will be forced to resolve merge conflicts before proceeding

moshenfeld:
Locally I don't see any indication of a conflict.
I'm seemingly fully synced with the remote version.
Is it possible I'm in the wrong branch or something similar, or perhaps
it's something only someone from within the organization can view?

ryan112358:
Can you try following these steps?

```
# 1. Add the upstream Google repository (if not done already)
git remote add upstream https://github.com/google/differential-privacy.git

# 2. Fetch the latest changes directly from Google's repository
git fetch upstream

# 3. Ensure you are on your feature branch
git checkout random-alocation

# 4. Merge the upstream main branch to trigger the conflicts locally
git merge upstream/main

# 5. Resolve the conflicts in your editor, then commit and push
git commit -m "Resolve merge conflicts with upstream main"
git push origin random-alocation
```

moshenfeld:
Done.
Seems like I had the Google repository up-streamed, I was on
the random-alocation branch, but the main upstream was not merged, and the
conflict came from a change that took place in main (I assumed all the code
changes in random allocation were still taking place in the branch).
It looks like like the delivery passed successfully and is now waiting for
maintainer approval.

arung54:
Thanks Moshe. I believe I don't have permission for maintainer approval, but have asked folks who work on OSS to help us proceed + also am working on the internal review. Will keep you posted

python/dp_accounting/dp_accounting/pld/random_allocation/BUILD.bazel
```diff
@@ -0,0 +1,156 @@
...
+        "@accounting_py_pip_deps//numpy",
+    ],
+)
+
+py_library(
+    name = "distributions",
+    srcs = ["distributions.py"],
+    srcs_version = "PY3",
+    deps = [
+        ":definitions",
```
RamSaw, line 50: Looking at [failing GitHub actions](https://github.com/google/differential-privacy/actions/runs/29639362112/job/88773038417?pr=414) looks like we need to add here dep on `'typing_extensions'`. Looking at how code is built I think you need to update [python/dp_accounting/requirements.txt](https://github.com/google/differential-privacy/blob/main/python/dp_accounting/requirements.txt) and https://github.com/google/differential-privacy/blob/main/python/dp_accounting/requirements_lock.txt

RamSaw (changes requested):
Hi! Thanks for contributing into our library! I will be working on merging this PR into Google3 internal codebase and making sure on GitHub changes are correctly attributed to you.

moshenfeld:
Thanks for the feedback.
I hope I got it right.

python/dp_accounting/dp_accounting/pld/privacy_loss_distribution_basic_example.py
arung54, line 1: Pritish suggested not touching this file since it is legacy and not meant to be exhaustive; can you remove from the PR?

moshenfeld: Done

python/dp_accounting/dp_accounting/pld/random_allocation/utils.py
```diff
@@ -0,0 +1,545 @@
...
+    # Kahan summation: compensated addition of current element
+    y = padded_probs[i] - compensation
+    t = running_sum + y
+    compensation = (t - running_sum) - y
+    running_sum = t
+
+  return ccdf
+
+
+def _ccdf_from_pmf(
```
arung54: Unused, can you remove?

python/dp_accounting/dp_accounting/pld/random_allocation/utils.py
```diff
@@ -0,0 +1,545 @@
...
+  expanded_pmf[indices] = pmf
+  return distributions.SparseDiscreteDist(
+      x_array=grid,
+      prob_arr=expanded_pmf,
+      p_min=dist.p_min,
+      p_max=dist.p_max,
+  )
+
+
+def _align_distributions_to_union_grid(
```
arung54: Unused, can you remove?

python/dp_accounting/dp_accounting/pld/random_allocation/utils.py
moshenfeld: Removed it as well as several other orphan functions.

python/dp_accounting/dp_accounting/pld/random_allocation/utils.py
moshenfeld: Removed it as well as several other orphan functions.

